PR #71, opened 2015-03-06
Various changes

IlyaSkriblovsky:
Hi!

I'm suggesting my changes made to txmongo while using it in production in my project.

The most important is 477f34a1, it fixes the following bug:
if find() is called with limit specified and the requested number of documents was returned by MongoDB in single batch, then to_fetch variable is set to zero and Getmore is called with n_to_return=0 which make MongoDB to return full next 16Mb-batch of documents although the requested number of them was already returned in the first batch. So, if you call find(limit=100) then MongoDB will return first batch with 100 documents, but also it will send the second unneeded batch with 16Mb of documents.
This bug was a huge performance issue for me.

Also, current implementation of find() doesn't close MongoDB cursors. To fix this, I also fixed KillCursors command serialization which was broken.

Also I've fixed 'hint' filter. It was incorrectly serialized with double dollar signs.

Best regards,
  Ilya


fiorix:
Waiting till others can review and if there are no objections I'll merge it in tomorrow.

psi29a:
Looks good so far, I'll go deeper tomorrow morning. 

@IlyaSkriblovsky Can you provide some unit testing to get some coverage on your changes please? :)

IlyaSkriblovsky:
OK, I will make some unit tests.
But I'm not sure how to test the main change — fix for find() method because my fix doesn't change it's external behavior, it just removes redundant getmore call to mongodb. I will try to find some way to mock up protocol methods to count how many requests to mongodb was sent.

IlyaSkriblovsky:
I've added tests that cover all my changes.

The most interesting are ones added to test_queries.py:
* test_CursorClosing() checks that there is no stale cursors after find() with specified limit
* test_GetMoreCount() is checking that no Getmore requests are sent to MongoDB if find() is called with small limit (this is done by ugly monkey-patching MongoClientProtocol method)

Both of them are failing on upstream master but passing on my branch.

psi29a:
That is great work! Thank you for the contribution.

fiorix:
@psi29a if you're happy go ahead and merge it.
Thanks @IlyaSkriblovsky! :)

PR #72, opened 2015-03-11
Authentication reimplemented to work well with ConnectionPool and with multiple DBs

IlyaSkriblovsky:
Hi!

The current implementation of authentication is broken in several use cases:

* It is not reenterant and this leads to bugs with ConnectionPool when pool_size > 1. In such case ConnectionPool.authenticate_with_nonce() might be called multiple times simultaneously and race-condition occurs because 'getnonce' and 'authenticate' MongoDB commands cannot be mixed up.

* It makes assumption that authentication on some DB is only required when accessing this DB. But MongoDB allows user from DB1 to have permissions on DB2.

I've reworked all auth-related stuff and added some unit tests, some of them are failing on master branch.

*Please note that unit tests in test_auth.py require your local MongoDB to have auth=true, no configured users and localhost exception enabled.*

### DIFF
--- a/tests/mongod.py
+++ b/tests/mongod.py
@@ -29,6 +29,7 @@ class Mongod(object):
                 '--dbpath', str(self.dbpath),
                 '--noprealloc', '--nojournal',
                 '--smallfiles', '--nssize', '1',
+                '--nohttpinterface',
                 ]
 
         if self.auth: args.append('--auth')
--- a/tests/mongod.py
+++ b/tests/mongod.py
@@ -1,0 +1,73 @@
+from twisted.internet import defer, reactor
+from twisted.internet.error import ProcessDone
+
+
+class Mongod(object):
+
+    # FIXME: this message might change in future versions of MongoDB
+    # but waiting for this message is faster than pinging tcp port
+    # so leaving this for now
+    success_message = 'waiting for connections on port'
+
+    def __init__(self, dbpath, port=27017, auth=False):
+        self.__proc = None
+        self.__notify_waiting = []
+        self.__notify_stop = []
+        self.__output = ''
+        self.__end_reason = None
+
+        self.dbpath = dbpath
+        self.port = port
+        self.auth = auth
+
+    def start(self):
+        d = defer.Deferred()
+        self.__notify_waiting.append(d)
+
+        args = ['mongod',
+                '--port', str(self.port),
+                '--dbpath', str(self.dbpath),
+                '--noprealloc', '--nojournal',
+                '--smallfiles', '--nssize', '1',
+                ]
+
+        if self.auth: args.append('--auth')
+
+        self.__proc = reactor.spawnProcess(self, 'mongod', args)
+        return d
+
+
+    def stop(self):
+        if self.__end_reason is None:
+            if self.__proc and self.__proc.pid:
+                d = defer.Deferred()
+                self.__notify_stop.append(d)
+                self.__proc.signalProcess('INT')
+                return d
+            else:
+                return defer.fail('Not started yet')
+        else:
+            if self.__end_reason.check(ProcessDone):
+                return defer.succeed(None)
+            else:
+                return defer.fail(self.__end_reason)
+
+    def makeConnection(self, process): pass
+    def childConnectionLost(self, child_fd): pass
+    def processExited(self, reason): pass
+
+    def childDataReceived(self, child_fd, data):
+        self.__output += data
+        if self.success_message in self.__output:
+            defs, self.__notify_waiting = self.__notify_waiting, []
+            for d in defs:
+                d.callback(None)
+
+    def processEnded(self, reason):
+        self.__end_reason = reason
+        defs, self.__notify_stop, self.__notify_waiting = self.__notify_stop + self.__notify_waiting, [], []
+        for d in defs:
+            if reason.check(ProcessDone):
+                d.callback(None)
+            else:
+                d.errback(reason)
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -28,6 +28,13 @@ mongo_uri = 'mongodb://{0}/'.format(mongo_host)
 
 
 class TestMongoAuth(unittest.TestCase):
+    """
+    NB: This testcase requires:
+        * auth=true in MongoDB configuration file
+        * no configured users
+        * localhost exception enabled (this is default)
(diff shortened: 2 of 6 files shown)


psi29a:
Should the 'note' also be included in the test_auth.py? We can add this as a changelog entry, but it would be nice to have it inline as well.

IlyaSkriblovsky:
Good point, I will include it today and also will try to add runtime check for auth=true in setUp()

IlyaSkriblovsky:
I've added runtime check for authentication enabled in Mongo and skipping auth tests if it is not.

Btw, do you know any proper way to ask Mongo whether it has auth enabled? I haven't found any. Currently I simply issue find_one and catch OperationFailure.

psi29a:
I've not come across anything either, so I think your try/catch is the best way to check for that.

I think any real meaningful test would require something like setUpClass to setup mongodb using subprocess Popen/Call and tearDownClass when finished. While running the test, you can issue commands and restart mongodb as necessary.

IlyaSkriblovsky:
I'm trying to make this testcase to start its own mongod process with --auth and --dbpath pointing to temporary folder. But I can't find anything like setUpClass or setUpModule in Trial. Only this ticket about setUpClass/tearDownClass was removed from Trial: https://twistedmatrix.com/trac/ticket/4175

May be you can advice me something? I can start/stop mongod before/after each test method, but this slows down testing and seems like overkill.

IlyaSkriblovsky:
Now test_auth.py starts separate 'mongod --auth' process for each test. With '--smallfiles --nssize 1' it runs in acceptable time. But if Trial has some support for setUpClass or setUpModule, please let me know. It would be much more efficient but I haven't found any.

Btw, may be it makes sense to run separate mongod for all txmongo's tests, not only test_auth.py, in order to not interfere with user's local mongo instance and to ensure there will be no traces left in database? However, not in this pull request of course.

psi29a:
I use setUpClass and tearDownClass (which sets things up for the entire life of the test run), but this requires pytest or nose. I use pytest with coverage at work where I setup a sharded/replicated 3 mongodb setup. I'm not sure if trial supports those or not, I've not tried.

However, it seems you've found a solution. How are the startup times? You can add --nohttpinterface to help speed things up since it isn't being used.

IlyaSkriblovsky:
Yeah, there is setUpClass/tearDownClass in standard pyunit, but they were removed from Trial somehow according to https://twistedmatrix.com/trac/ticket/4175 . Don't know why.

All 6 tests from test_auth.py take ~2 seconds on my desktop.

I've added --nohttpinterface, but it only has effect on MongoDB ≤2.4 since HTTP is disabled by default on 2.6+

psi29a:
I didn't know that, thanks! :)

Does this also mean that with CI integration with appveyor and travis, that we'll need to install mongodb during testing (just by running tox or trial tests) by default?

IlyaSkriblovsky:
I'm not familiar with CI tools, but I don't think anything is changed because other unit tests already expects mongodb to be listening on localhost:27017, so mongodb already must be installed

IlyaSkriblovsky:
I've found that test_auth.py fails on MongoDB 3.0 because txmongo only supports old MONGODB-CR authentication protocol and also does my patch. But 3.0 doesn't allow new users to be authenticated with MONGODB-CR.

Note that my patch doesn't make anything worse, txmongo simply had no tests for authentication before.

Should I just make test_auth.py to skip its tests on 3.0 until someone (may be I) will contribute SCRAM-SHA-1 authentication to txmongo?

psi29a:
We can fail for now with 3.0, no need to go nuts on version checking.  I'm surprised we've gotten as far as we've had. :) If you want to do it, and have time please do so... in another pull request. :) We'll need SCRAM-SHA-1 at some point.

I have a patchset I'm working on to get txmongo's index_information more inline with pymongo's. Right now it doesn't return back in any information about sparse and unique indexes.  Would you be upset if we used pymongo's style of just returning a list or would you prefer a dict with the index name as the key like we currently do?

IlyaSkriblovsky:
Ok lets auth tests fail on 3.0 for now. So this pull request is ready for review.

It seems like in PyMongo 1.7+ index_information() also returns dict with unique and sparse flags: http://api.mongodb.org/python/current/api/pymongo/collection.html#pymongo.collection.Collection.index_information
Receord in Changelog: http://api.mongodb.org/python/current/changelog.html#changes-in-version-1-7

But I'm personally don't use index_information() in my projects and doesn't mind against any changes to it :)

psi29a:
I use it, to try to validate my indexes are in place before trying to add them... again. ;)  There is a test for it, but it commented out due to the fact that it isn't currently supported because of a format change in MongoDB 1.x  I'm of the opinion that txmongo should try to stay close to pymongo, just to keep it sane/simple for us, but it will be breakage for those that depend on a dict and not a list.

I'll give this another look through then merge. :)

IlyaSkriblovsky:
I agree that it's good for txmongo to be compatible in API with pymongo as far as possible

But i'm didn't caught up what you mean by "pymongo's style of just returning a list" because PyMongo also returns dict, not  (but of another form than current txmongo).

psi29a:
You're right, the documentation says dict. http://api.mongodb.org/python/current/api/pymongo/collection.html

However, when we ask over the wire from txmongo to mongodb itself, mongodb gives us an plain unordered list.  It is my intent to mimic pymongo, not mongodb itself.

I'll try to get my patch set ready, please review when you get the chance. :)

IlyaSkriblovsky:
Aaah, now I see what you mean. Sure my vote is for pymongo's style, although it's less pretty. Compatibility is more important, i think. I'll be happy to review your patch when it's ready.

fiorix:
FYI http://docs.travis-ci.com/user/database-setup/#MongoDB

IlyaSkriblovsky:
So, Travis starts MongoDB with auth enabled by default, cool. But does current configuration of Travis has MongoDB configured? I don't see any mention of MongoDB in .travis.yml. May be I'm missing something because I haven't worked with Travis before

psi29a:
I'll take care of it when we finally get this running on Travis, if it doesn't get installed we'll make sure it does and has at least one instance running. :)

psi29a:
I've also had to do a rebase, a lot of things has changed so there was some conflict. Got that cleaned up and now updating news about your recent contribs.

PR #73, opened 2015-03-12
Index information now returns all information

psi29a:
- Return more in index_information, using a dict like pymongo
- Replaced xrange with range to get things future proof for python3
- pep8 as much as possible, 100 column limit
- Fixed errnos and messages
- Use " for strings and ' for single chars, just to be consistent
- Updated NEWS

### DIFF
--- a/tests/test_collection.py
+++ b/tests/test_collection.py
@@ -171,13 +171,13 @@ class TestIndexInfo(unittest.TestCase):
         yield db.test.create_index(filter.sort(filter.ASCENDING("hello")))
         ix_info = yield db.test.index_information()
         self.assertEqual(len(ix_info), 2)
-        self.assertIsNotNone(ix_info["hello_1"]["name"] == "hello_1")
+        self.assertEqual(ix_info["hello_1"]["name"], "hello_1")
 
         yield db.test.create_index(filter.sort(filter.DESCENDING("hello") +
                                                filter.ASCENDING("world")),
                                    unique=True, sparse=True)
         ix_info = yield db.test.index_information()
-        self.assertIsNotNone(ix_info["hello_1"]["name"] == "hello_1")
+        self.assertEqual(ix_info["hello_1"]["name"], "hello_1")
         self.assertEqual(len(ix_info), 3)
         self.assertEqual({"hello": -1, "world": 1}, ix_info["hello_-1_world_1"]["key"])
         self.assertEqual(True, ix_info["hello_-1_world_1"]["unique"])
--- a/tests/test_collection.py
+++ b/tests/test_collection.py
@@ -1,6 +1,4 @@
-# -*- coding: utf-8 -*-
-
-# Copyright 2012 Renzo S.
+# Copyright 2012-2015 TxMongo Developers
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -33,7 +31,7 @@ mongo_host = "localhost"
 mongo_port = 27017
 
 
-class TestCollection(unittest.TestCase):
+class TestIndexInfo(unittest.TestCase):
     
     timeout = 5
 
@@ -45,8 +43,8 @@ class TestCollection(unittest.TestCase):
 
     @defer.inlineCallbacks
     def tearDown(self):
-        ret = yield self.coll.drop()
-        ret = yield self.conn.disconnect()
+        yield self.coll.drop()
+        yield self.conn.disconnect()
 
     @defer.inlineCallbacks
     def test_collection(self):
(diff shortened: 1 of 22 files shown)


psi29a:
My apologies for such a large PR, it started out as just fixing index_information but it also contains a lot of pep8, nit-pickyness and others bits to try to get the code-base to be consistent.

I try to adhere to the 100 column length, as 80 as specified by pep8 is unrealistic. We should have eventually a contributions guideline but that will come later.

fiorix:
It may seem unrealistic but 100 makes no sense. The reason for having 80 columns (as the linux kernel coding style specifies, Go fmt and others) is because it works for everyone, especially when you're in front of a console (e.g. vnc console of your cloud provider, openstack, etc) and your code will look bad and unusable.

fiorix:
I'd recommend to stick to 80 as pep8 mandates, although not very strictly - strings, error message etc can be an exception, use your best judgement. For for general code, 80 is more appropriate.

psi29a:
pep8 is a nice guideline.

Thankfully there wasn't much formatting necessary since most of it was already formatting for 80. As you said however, there where moments that the wasted white-space by bring it to the next line with strings and other bits would make it harder to read and we never got over 100 which is why that magic number was chosen.

80 characters makes sense when you are using 320x200 CRT (4:3), but most of us use 16:9 and have a framebuffer that can handle it.  I would go so far as say 120 is ideal, vim is happy and looks good in my console. :)

Generally, for code... if you are pushing 80, you're  likely doing it wrong. ;)

tests/test_collection.py
```diff
@@ -169,51 +166,47 @@ def test_index_info(self):
...
-
-        self.assert_("_id_" in ix_info)
+        self.assertEqual(ix_info["_id_"]["name"], "_id_")
 
         yield db.test.create_index(filter.sort(filter.ASCENDING("hello")))
         ix_info = yield db.test.index_information()
         self.assertEqual(len(ix_info), 2)
-        
-        self.assertEqual(ix_info["hello_1"], [("hello", 1)])
+        self.assertIsNotNone(ix_info["hello_1"]["name"] == "hello_1")
```
IlyaSkriblovsky: `assertIsNotNone(a == b)` will always pass since both True and False are not None :)

tests/test_collection.py
```diff
@@ -169,51 +166,47 @@ def test_index_info(self):
...
 
-        yield db.test.create_index(filter.sort(filter.DESCENDING("hello") + filter.ASCENDING("world")), unique=True)
+        yield db.test.create_index(filter.sort(filter.DESCENDING("hello") +
+                                               filter.ASCENDING("world")),
+                                   unique=True, sparse=True)
         ix_info = yield db.test.index_information()
-
-        self.assertEqual(ix_info["hello_1"], [("hello", 1)])
+        self.assertIsNotNone(ix_info["hello_1"]["name"] == "hello_1")
```
IlyaSkriblovsky: Same as above: `assertIsNotNone()` should be probably `assertTrue()`

tests/test_objects.py
```diff
@@ -166,37 +163,32 @@ def test_GridFsOperations(self):
...
             yield g_in.close()
             
             # Tests reading from an existing gridfs file
-            g_out = yield gfs.get_last_version('optest')
+            g_out = yield gfs.get_last_version("optest")
             data = yield g_out.read()
             out_file.write(data)
             _id = g_out._id
-        except Exception,e:
+        except Exception, e:
```
IlyaSkriblovsky: Since this commit is mainly related to coding style, may be its worth to change `except Exception, e:` to `except Exception as e:`?

tests/test_queries.py
```diff
@@ -67,66 +67,66 @@ def test_FindWithCursor(self):
...
 
         # Calculate number of objects in 4mb batch
-        obj_count_4mb = 4 * 1024**2 / len(BSON.encode(makeobj())) + 1
+        obj_count_4mb = 4 * 1024**2 / len(BSON.encode(make_object())) + 1
 
         first_batch = 5
-        yield self.coll.insert([makeobj() for i in xrange(first_batch + obj_count_4mb)])
-        result = yield self.coll.find(limit = first_batch)
+        yield self.coll.insert([make_object() for _ in range(first_batch + obj_count_4mb)])
+        _ = yield self.coll.find(limit=first_batch)
```
IlyaSkriblovsky: You may safely remove `_ = ` here

IlyaSkriblovsky:
You didn't changed quotes in KillCursors class and Msg namedtuple in protocol.py :)  Probably because you've merged my changes after making yours

psi29a:
Yeah, I had to do a rather nasty rebase with merge conflicts. ;)

I'll take a good look in a bit. Thanks for eyeballing this.

psi29a:
Another rebase after the last merge request, this time also fixing up the issues eyed by @IlyaSkriblovsky 

Cheers!

IlyaSkriblovsky:
I'm happy with this changeset

psi29a:
Merging this in so as to not be a blocker for others.  If there is anything I've missed, please comment and I'll open a new PR to fix/change it.

PR #77, opened 2015-03-14
SCRAM-SHA-1 authentication

IlyaSkriblovsky:
Support for SCRAM-SHA-1 authentication protocol. It was introduced in MongoDB 2.8 and became default in 3.0+.

`ConnectionPool.authenticate()` now accepts `mechanism` optional argument that might be used to force MONGODB-CR auth protocol with MongoDB 3.0+. Default is to autodetect protocol from server wire protocol version.

### DIFF
--- a/txmongo/connection.py
+++ b/txmongo/connection.py
@@ -40,14 +40,16 @@ class _Connection(ReconnectingClientFactory):
         # Build the protocol.
         p = ReconnectingClientFactory.buildProtocol(self, addr)
 
+        ready_deferred = p.connectionReady()
+
         if not self.uri['options'].get('slaveok', False):
             # Update our server configuration. This may disconnect if the node
             # is not a master.
-            p.connectionReady().addCallback(lambda _: self.configure(p))
+            ready_deferred.addCallback(lambda _: self.configure(p))
 
-        p.connectionReady()\
+        ready_deferred\
             .addCallback(lambda _: self._auth_proto(p))\
-            .addCallback(lambda _: self.setInstance(instance=p))
+            .addBoth(lambda _: self.setInstance(instance=p))
         return p
 
     def configure(self, proto):
@@ -98,6 +100,9 @@ class _Connection(ReconnectingClientFactory):
         max_bson_size = config.get("maxBsonObjectSize")
         if max_bson_size:
             proto.max_bson_size = max_bson_size
+
+        proto.set_wire_versions(config.get("minWireVersion", 0),
+                                config.get("maxWireVersion", 0))
 
         # Track the other hosts in the replica set.
         hosts = config.get("hosts")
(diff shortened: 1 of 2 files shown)


IlyaSkriblovsky:
This code uses various private functions from `pymongo.auth`. What do you think about this practice?
New version of PyMongo may break compatibility at any time (although unlikely). But duplicating these functions from PyMongo doesn't looks good to

(By the way, txmongo's authentication code already used `pymongo.auth._auth_key` long before any my changes)

psi29a:
Before we merge this, I would advise that we first merge my branch and that you rebase. Otherwise, again, great work. :)

IlyaSkriblovsky:
I did the rebase of my changes on top of the current master

IlyaSkriblovsky:
How about this one?

psi29a:
Just got in, it looks good but I can't test it at the moment.  Is there anyway to get some coverage here? Is MongoDB3 backward compatible with older methods? Is there a way to get version information out of MongoDB?

IlyaSkriblovsky:
This implementation does same as current PyMongo does. It looks on [maxWireVersion field of isMaster command] (http://docs.mongodb.org/manual/reference/command/isMaster/#isMaster.maxWireVersion) and decides which auth method to use (this can be overwritten by `mechanism='MONGODB-CR'` if needed).

`test_auth.py` starts distinct mongod instance for each test and creates users at runtime. But there is no way to create MONGODB-CR users in v3.0¹ or SCRAM-SHA-1 users in v2.x. So it seems impossible to test both auth mechanisms on a single host. Actual auth method that will be tested by `test_auth.py` is choosen at runtime depending on which MongoDB version you have installed.

¹ MongoDB v3.0 is compatible with MONGODB-CR but only for existing user accounts which were created by v2.x.

psi29a:
That is real pity that we are in an exclusive or position considering that most platforms will not come with two mongodb binaries. So we'll never be able to reach the theoretical 100% coverage in one run.

IlyaSkriblovsky:
Yeah, but we are not worse here than official PyMongo that doesn't seems to test MONGODB-CR on v3.0
https://github.com/mongodb/mongo-python-driver/blob/master/test/test_auth.py

May be it would be more honest to add two tests: `test_MONGODB_CR` and `test_SCRAM_SHA_1` that will test specific auth mechanisms if they are supported by current MongoDB installation and will be skipped otherwise. This would not help coverage, but will make absense of 100% coverage visible to person running tests by reporting that some tests was skipped.

Another crazy idea is to generate minimal database with MONGODB-CR users using MongoDB v2.x and to place it as blob into TxMongo's code tree. Then testing code can run mongod against these files and test both MONGODB-CR and SCRAM-SHA-1 thanks to backward compatibility for existing users (assuming we have v3.0 installed).

psi29a:
While I'm a big fan of crazy, I rather not pollute the repo. Another solution would be to have travis-ci to pull in custom built versions of mongodb (v2 and v3) but I think this is rather overkill.

So long as we can run the tests manually, with the understanding that some parts of version dependent is good enough for us developers.

I think at this point, we should just document it. What do you think if I (or you) create a new .rst file explaining the new MongoDB3 support with what you've already written here. That way it will show up in readthedocs.org and we can link to in the changelog.

Right, going to merge this unless you (or anyone) has any objections. Again, thank you! :)

IlyaSkriblovsky:
> What do you think if I (or you) create a new .rst file

May I ask you to do this? I'm not familiar (yet) with python documentation infrastructure and its good practices.

psi29a:
Sure, no worries.

PR #78, opened 2015-03-15
find() as wrapper around fixed find_with_cursor()

IlyaSkriblovsky:
I have fixed same problems with `limit=` in `find_with_cursor()` as 477f34a1190b272265f1702e3e11875a58457125 fixed for `find()`

After that I noticed that `find()` and `find_with_cursor()` are almost identical while `find_with_cursor()` is the superset of `find()`. So, 539ca559b9be06715f9094c61da45cb068601538 replaces `find()` implementation with simple code that unwinds result of `find_with_cursor()`.

I don't see any way compatibility or something might be broken because they do literally the same, including error-handling (i.e. absense of it :) ).

### DIFF
--- a/tests/test_queries.py
+++ b/tests/test_queries.py
@@ -76,6 +76,25 @@ class TestMongoQueries(unittest.TestCase):
         self.assertEqual(total, 750)
 
     @defer.inlineCallbacks
+    def test_FindWithCursorLimit(self):
+        yield self.coll.insert([{'v': i} for i in range(750)], safe=True)
+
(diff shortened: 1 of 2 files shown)


psi29a:
Nice work, less code the better!

I'll give it a good/long look in a bit. 

At this rate, we might have a new release soon. I'll also ask @glyph about getting this repo moved over to Twisted proper.

trenton42:
+1 to getting this moved over to twisted. It got hard for a while to track which fork was the active one

psi29a:
Just read through it, I don't have any comments. Great work and nice inline comments!

Issue #80, opened 2015-04-14
Compatibility issues with pymongo3 & bson3


boogheta:
Hi, using txmongo for a while I hadn't updated for about a year and am discovering now the new versions.

I tried installing on a blank ubuntu machine both mongo3 (following new method given [here](http://docs.mongodb.org/manual/tutorial/install-mongodb-on-ubuntu/)) then `pip install txmongo==0.6` and resulted in a non-functional install, as shown by the following error when running `python examples/inlinecallbacks/query.py`: 

<pre>
  File "/home/boo/.virtualenvs/hyphe/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 424, in errback
    self._startRunCallbacks(fail)
  File "/home/boo/.virtualenvs/hyphe/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 491, in _startRunCallbacks
    self._runCallbacks()
  File "/home/boo/.virtualenvs/hyphe/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 578, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/home/boo/.virtualenvs/hyphe/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1163, in gotResult
    _inlineCallbacks(r, g, deferred)
--- <exception caught here> ---
  File "/home/boo/.virtualenvs/hyphe/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1105, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "/home/boo/.virtualenvs/hyphe/local/lib/python2.7/site-packages/twisted/python/failure.py", line 389, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "/tmp/test", line 7, in example
    docs = yield mongo.foo.test.find(limit=10)
  File "/home/boo/.virtualenvs/hyphe/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1107, in _inlineCallbacks
    result = g.send(result)
  File "/home/boo/.virtualenvs/hyphe/local/lib/python2.7/site-packages/txmongo/collection.py", line 150, in find
    defer.returnValue([d.decode(as_class=as_class) for d in documents])
exceptions.TypeError: decode() got an unexpected keyword argument 'as_class'
</pre>

[...]

PR #81, opened 2015-04-14
pin pymongo 2.7 until bson3 deprecations handled

boogheta:
see #80


psi29a:
Fixed in https://github.com/twisted/txmongo/pull/82

PR #82, opened 2015-04-14
Fixed 'as_class' argument of find() and find_one() with pymongo 3.0

IlyaSkriblovsky:
PyMongo 3.0 removed `as_class` argument from BSON.decode as stated in #80
This fix adapts txmongo to this change by using new `CodecOptions` class from PyMongo 3.0 while falling back to `as_class` on 2.x.
Also testcase added for checking `as_class` behavior.

### DIFF
--- a/tests/test_queries.py
+++ b/tests/test_queries.py
@@ -18,6 +18,7 @@ from twisted.internet import defer
 from twisted.trial import unittest
 import txmongo
 from txmongo.protocol import MongoClientProtocol
+from collections import OrderedDict
 
 mongo_host = "localhost"
 mongo_port = 27017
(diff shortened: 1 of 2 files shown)


IlyaSkriblovsky:
Not sure that `if pymongo.version >= '3'` is the best way to checking for version of PyMongo.

May be try'ing to import CodecOptions and setting global variable, say, `_USING_CODECOPTIONS=True` in case of success? Doesn't looks much nicier to me.

Please advice if you know better way

psi29a:
Glad you found a workaround!  I honestly think it best to try/except CodecOptions, because I don't trust pymongo.version to always follow the same version scheme.

There should also be a inline comment explaining why the try/except is necessary in the code area.

psi29a:
Does bson.codec_options exist in pymongo < 3 ? If so, then we don't need the extra try/except at the top for imports.

psi29a:
At the top, we can just do import bson and later in the code use the try/except :
options = bson.codec_options.CodecOptions(document_class=as_class)

I think we can make a exception here to avoid having to try/except in the includes at the top. We also make it explicit/clear in the code that this is special.

IlyaSkriblovsky:
Thanks, I've rewrote it using `try ... except` and catching AttributeError

psi29a:
Awesome! :D

This is certainly 15.0 release material. ;)

PR #83, opened 2015-05-12
Named indexes

psi29a:
1) request for comments
2) testing for coveralls

This should fix issue:
https://jira.mongodb.org/browse/SERVER-18411

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: python
 python: 2.7
+services:
+    - mongodb
 
 env:
     - TOX_ENV=py26-twlatest
--- a/tests/test_filters.py
+++ b/tests/test_filters.py
@@ -39,6 +39,12 @@ class TestMongoFilters(unittest.TestCase):
 
     @defer.inlineCallbacks
     def test_Hint(self):
+        # Ensure there is no {x:1} index
+        yield self.coll.dropIndex([('x', 1)])
+
+        # find() should fail with 'bad hint' if hint specifier works correctly
+        self.assertFailure(self.coll.find({}, filter=qf.hint([('x', 1)])), OperationFailure)
+
         # create index and test it is honoured
         yield self.coll.create_index(qf.sort(qf.ASCENDING("x")), name="test_index")
         found_1 = yield self.coll.find({}, filter=qf.hint([('x', 1)]))
@@ -50,12 +56,6 @@ class TestMongoFilters(unittest.TestCase):
         self.assertFailure(self.coll.find({}, filter=qf.hint(["test_index", 1])), OperationFailure)
         self.assertFailure(self.coll.find({}, filter=qf.hint(qf.ASCENDING("test_index"))),
                            OperationFailure)
-
-        # Ensure there is no {x:1} index
-        yield self.coll.dropIndex([('x', 1)])
-
-        # find() should fail with 'bad hint' if hint specifier works correctly
-        self.assertFailure(self.coll.find({}, filter=qf.hint([('x', 1)])), OperationFailure)
 
     @defer.inlineCallbacks
     def test_Comment(self):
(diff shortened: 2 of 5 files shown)


psi29a:
We need someone to hook up travis and coveralls please!
@glyph help?

IlyaSkriblovsky:
Ouch. TestMongoFilters.test_Hint seems to be failing for me on MongoDB 3.0.3.

Failing line is the last assertFailure in test_Hint which is from my older commit d0c7ade9108e6c533a69997a2eefb5fb95525784. It breaks because `dropIndex` in previous line doesn't work... because it completely wrong! It should be `yield self.coll.drop_index(qf.sort(qf.ASCENDING('x')))`. But this is wrong too because `drop_index` can't delete named indexes and after your changes it should be written as `yield self.coll.drop_index('test_index')`.

Sorry, this is mostly my fault. Before your changed my `dropIndex` line was only for documenting the intention to run `find()` with hint of absent index. This index wasn't exist anyway, so wrong `dropIndex` was not causing problems.

(Btw, why drop_index issues `deleteIndexes` command while it is named `dropIndexes` in docs: http://docs.mongodb.org/manual/reference/command/dropIndexes/#dbcmd.dropIndexes ? But it works somehow).

IlyaSkriblovsky:
Aren't there same issue with named index support of `sort`? 

I must also admit that checking for `operation == "hint"` doesn't looks nice in `_QueryFilter._index_document` because in this way base class makes check for specific child class. Same check in `Collection.__apply_find_filter` also looks hacky.

Outside of scope of this particular pull request, I would suggest to move `Collection.__apply_find_filter` method to `_QueryFilter` since it does some checks which are specific to internal representation of filters.

psi29a:
Try again after another pull, I've been modifying the code as I go. ;)

I too noticed that drop_index isn't working, I pushed this to the top for the time being.

psi29a:
@IlyaSkriblovsky  If you would like to try it out in an apart branch (from master), please do.  I'm currently fighting travisci/tox/coveralls.

IlyaSkriblovsky:
BTW, `types.ListType` and `types.TupleType` are not Python3-ready. It would be more portable to check with `isinstance(x, (tuple, list))` (and more readable too).

IlyaSkriblovsky:
I've created #84 with simplier solution for original issue. Please take a look. Am I missing something? This way we don't need special checks for `hint` in `_index_document` and `__apply_find_filters`. It passes your new tests on both 2.6 and 3.0.3.

psi29a:
Be careful, because I tested this:  isinstance(x, (tuple, list))

(tuple, list) == tuple or   isinstance((tuple,list), tuple) is always true

IlyaSkriblovsky:
Sorry, didn't caught a point.

`isinstance` allows second argument be a type of tuple of types. So, `isinstance(x, (tuple, list))` checks whether `x` is `tuple` or `list`. Could you provide an example when it's wrong?

```
>>> import types
>>> types.TupleType is tuple
True
>>> types.ListType is list
True
```

psi29a:
Nothing wrong per say,  but look closely to this:
In [5]: isinstance(test, (list, tuple))
Out[5]: True

In [6]: isinstance(test, tuple)
Out[6]: True

why (list, tuple) and not just tuple, because (list, tuple) will resolve to base type tuple anyway in isinstance.

psi29a:
Oh! I didn't know you could chain it to check for both, I just learned something new. Thank you! :)

IlyaSkriblovsky:
:)

Also, `isinstance` doesn't treat second argument as "instance of type to check subclass of". So, `isinstance( (1,2), (int, str) ) == False` event while second argument is `tuple` and first argument is `tuple`. Just like `isinstance(5, 7)` won't check that 5 is int, but throw `TypeError` instead

PR #87, opened 2015-05-13
OrderedDict replaced by custom dict subclass for the sake of py26

IlyaSkriblovsky:
Python 2.6 doesn't have OrderedDict, so I replaced it with custom one-line `dict` subclass for testing `as_class` parameter of `find()`

### DIFF
--- a/tests/test_aggregate.py
+++ b/tests/test_aggregate.py
@@ -50,8 +50,8 @@ class TestAggregate(unittest.TestCase):
 
         res = yield self.coll.aggregate([{"$match": {"oh": "hai"}}], full_response=True)
 
-        self.assertIn("ok", res)
-        self.assertIn("result", res)
+        self.assertTrue("ok" in res)
+        self.assertTrue("result" in res)
         self.assertEqual(len(res["result"]), 2)
 
     @defer.inlineCallbacks
--- a/tests/test_collection.py
+++ b/tests/test_collection.py
@@ -86,7 +86,7 @@ class TestIndexInfo(unittest.TestCase):
         self.assertEqual(self.db.test.test, self.db.test("test"))
 
         options = yield self.db.test.options()
-        self.assertIsInstance(options, dict)
+        self.assertTrue(isinstance(options, dict))
 
         yield self.db.drop_collection("test")
         collection_names = yield self.db.collection_names()
--- a/tests/test_queries.py
+++ b/tests/test_queries.py
@@ -18,7 +18,6 @@ from twisted.internet import defer
 from twisted.trial import unittest
 import txmongo
 from txmongo.protocol import MongoClientProtocol
-from collections import OrderedDict
 
 mongo_host = "localhost"
 mongo_port = 27017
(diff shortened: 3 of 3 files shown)


psi29a:
Let's see what travis says. :)

It is a pity that we can't use OrderedDict, should we think about making an compat.py for these kinds of things? Or for this case, it isn't even necessary?

IlyaSkriblovsky:
We don't need OrderedDict in txmongo, it was used only for testing if `find(as_class=...)` works correctly.

Travis complaints, but not on OrderedDict anymore, so this PR might be accepted

IlyaSkriblovsky:
Ouch, I've noticed in Travis that test_Cursor* tests are failing when parsing response of `serverStatus` MongoDB command. What version of MongoDB is used by Travis?

psi29a:
According to this: https://github.com/travis-ci/travis-ci/issues/2246
they are using 2.4, which we don't support. We should only be targeting 2.6 and 3.0 .

I'll whip something up.

psi29a:
MongoDB version
MongoDB 2.4.12

Right, then I'll disable travis's mongodb and install 3.0 since that is what we are targetting. It will be nice to test 2.6, but I want the forward momentum.  What do you think?

psi29a:
I would go with 3.0.3 but it seems that the repo doesn't have it yet:
http://downloads-distro.mongodb.org/repo/ubuntu-upstart/dists/dist/10gen/binary-amd64/

Are we good with 2.6.9 ?

IlyaSkriblovsky:
Sure, but we won't test SCRAM-SHA-1 on 2.6.x. But with 3.0 we won't test MONGODB-CR :)
I think it is better to test on 2.6 since it is most commonly used.

Btw, I've pushed #89 that fixes `serverStatus` on MongoDB 2.4, it should be ok now

psi29a:
Can you rebase against master and force push?  That should trigger travis again, hopefully we're green. :)

psi29a:
Awesome, we finally got some greens and coveralls support! :) 74% isn't bad, but we should aim for 80%

psi29a:
[ERROR]
  File "/home/travis/build/twisted/txmongo/.tox/py26-tw130/lib/python2.6/site-packages/twisted/internet/defer.py", line 1070, in _inlineCallbacks
    result = g.send(result)
  File "/home/travis/build/twisted/txmongo/tests/test_queries.py", line 213, in test_AsClass
    self.assertIs(type(doc), dict)
exceptions.AttributeError: 'TestMongoQueries' object has no attribute 'assertIs'
tests.test_queries.TestMongoQueries.test_AsClass

psi29a:
Can we replace the self.assertIs with self.assertTrue ?

IlyaSkriblovsky:
Yep, please see b142fa1091e19cbfd4b57bede633a57386b2f83d that i've pushed just now

psi29a:
beautiful, we're almost there :)
thanks for working with me on this.

IlyaSkriblovsky:
Thanks for the infrastructure support you've implemented! It would help a lot to keep code testable and bulletproof

PR #88, opened 2015-05-13
we cannot guarentee the order of a list, so we make sure all elements…

psi29a:
… are find in a list

This is to fix this from happening:
https://travis-ci.org/psi29a/txmongo/jobs/62363503

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: python
 python: 2.7
-services:
-    - mongodb
 
 env:
     - TOX_ENV=py26-twlatest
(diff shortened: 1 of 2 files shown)


psi29a:
@IlyaSkriblovsky you agree with this fix?

IlyaSkriblovsky:
Sure, it's good to go

psi29a:
I added support for the latest official mongodb from their ubuntu archives... currently that is version 2.6.9
I don't know when 3.0.x will go in, hopefully soon.

psi29a:
Travis seems to be happy now, the only failing test I see is about the orderedict.

PR #89, opened 2015-05-13
Support for old-style serverStatus command on MongoDB 2.4

IlyaSkriblovsky:
Falling back to `serverStatus.cursors.totalOpen` instead of `serverStatus.metrics.cursor.open.total` when we are on MongoDB 2.4

PR #90, opened 2015-05-13
Support for old-style serverStatus command on MongoDB 2.4

IlyaSkriblovsky:
Falling back to `serverStatus.cursors.totalOpen` instead of `serverStatus.metrics.cursor.open.total` when we are on MongoDB 2.4


IlyaSkriblovsky:
This one is instead #89

PR #91, opened 2015-05-13
__builtins__["len"] -> __builtins__.len for PyPy

IlyaSkriblovsky:
Somewhy I was using subscription syntax for __builtins__ module instead of dot syntax (don't know why :)). This is not supported by PyPy.


IlyaSkriblovsky:
Stupid mistake: I was using `__builtins__` instead of `__builtin__` module :frowning:

psi29a:
Looks like we're almost there. :)

PR #92, opened 2015-05-13
Tests that encode and decode protocol messages and compare results

IlyaSkriblovsky:
A bunch of tests that encode protocol messages into bytes and then decode them using `txmongo.protocol.MongoDecoder` and compare decoded object with original request object.

This might be considered as additional check for encoding code. Also this will add test coverage for `MongoDecoder` class.

### DIFF
--- a/tests/test_protocol.py
+++ b/tests/test_protocol.py
@@ -37,13 +37,11 @@ class _FakeTransport(object):
 
 class TestMongoProtocol(unittest.TestCase):
 
-    def __test_encode_decode(self, method, request):
-        # `method` is unbound send_* method of MongoClientProtocol
-
+    def __test_encode_decode(self, request):
         proto = MongoClientProtocol()
         proto.transport = _FakeTransport()
 
-        method(proto, request)
+        proto.send(request)
 
         decoder = MongoDecoder()
         decoder.feed(proto.transport.get_content())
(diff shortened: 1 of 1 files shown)


IlyaSkriblovsky:
Failing tests in Travis are not linked to this PR. They are failings with "reactor unclean" and "trying to cancel already cancelled event" that sporadically appear on various tests from time to time (there is possbile race condition somewhere).

psi29a:
I saw, I can restart the tests and they eventually are green.

It happens on py26,27,pypy in almost all twisted versions.... so it will be a problem to track down.
Nice coverage improvement btw! :)

psi29a:
I just restarted the 'whole' build process to be sure, and all builds are green. It might have something to do with load on the build machines?

IlyaSkriblovsky:
I saw same strange fails on my own PC. But fail probability was higher when
I ran tests in VirtualBox, so the probability seems to be somehow linked to
machine performance.

I bet txmongo (or its unittests) doesn't wait correctly for all cleanups.
And on showers machines cleanup takes a bit longer and trial complains
about dirty reactor.

BTW, these fails are always (or may be almost always) appear on aggregate
and gridfs tests.

I want to look into this issue today or tomorrow evening.

18:13, чт, 14.05.2015, Bret Curtis <notifications@github.com>:

psi29a:
Sure, i'm doing some additional coverage in the meantime.

PR #93, opened 2015-05-14
add create_collection coverage

psi29a:
drop collection via database

cover drop collection typeerror

### DIFF
--- a/tests/test_objects.py
+++ b/tests/test_objects.py
@@ -37,8 +37,17 @@ class TestMongoObjects(unittest.TestCase):
         conn = yield txmongo.MongoConnection(mongo_host, mongo_port)
         mydb = conn.mydb
         self.assertEqual(isinstance(mydb, database.Database), True)
+        self.assertEqual(repr(mydb), "Database(Connection('localhost', 27017), u'mydb')")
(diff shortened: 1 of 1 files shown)


psi29a:
We just broke 80% coverage! :)

trenton42:
@psi29a Way to go! I really appreciate everything you all are doing here!

psi29a:
Give a big thanks to @IlyaSkriblovsky as he has been really on the ball.  :)

IlyaSkriblovsky:
Cool!

I've looked into sporadically failed tests and found that this happens for tests that create `MongoConnection` and immediately call its `disconnect()` without issuing any queries. `MongoConnection()` constructor starts connecting in background. Connection process in turn starts name resolution of "localhost" into ip-address. If name resolution doesn't finish until `disconnect()` returns (it takes exactly 1 main loop round-trip), trial fails with "reactor unclean". (Name resolution is deferred to separate thread, so there is race condition which fails more often on slow machines).

Unfortunately, I didn't found any good way to wait until name resolution is done :(

The simpliest way to solve this is to change "localhost" to "127.0.0.1" in `test_queries.py` and `test_aggregate.py`. Or to issue any command on `MongoConnection` in corresponding test methods.

PR #94, opened 2015-05-14
Trying to fix "reactor unclean" test fails

IlyaSkriblovsky:
I've looked into sporadically failed tests and found that this happens for tests that create MongoConnection and immediately call its disconnect() without issuing any queries. MongoConnection() constructor starts connecting in background. Connection process in turn starts name resolution of "localhost" into ip-address. If name resolution doesn't finish until disconnect() returns (it takes exactly 1 main loop round-trip), trial fails with "reactor unclean". (Name resolution is deferred to separate thread, so there is race condition which fails more often on slow machines).

Unfortunately, I didn't found any good way to wait until name resolution is done :(

The simpliest way to solve this is to change "localhost" to "127.0.0.1" in test_queries.py, test_aggregate.py and test_connection.py. Or to issue any command on MongoConnection in corresponding test methods.

After changing "localhost" to "127.0.0.1" I'm not able to reproduce this anymore.

Lets see what Travis will say

### DIFF
--- a/tests/test_aggregate.py
+++ b/tests/test_aggregate.py
@@ -17,7 +17,7 @@ from twisted.internet import defer
 from twisted.trial import unittest
 import txmongo
 
-mongo_host = "localhost"
+mongo_host = "127.0.0.1"
 mongo_port = 27017
 
 
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -17,7 +17,7 @@ from twisted.trial import unittest
 from twisted.internet import base, defer
 import txmongo
 
-mongo_host = "localhost"
+mongo_host = "127.0.0.1"
 mongo_port = 27017
 base.DelayedCall.debug = True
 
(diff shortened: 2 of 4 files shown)


IlyaSkriblovsky:
Ouch. This is another fail, not covered by this PR. This TimeoutError looks like solely by low performance of Travis' VM.

Is it possible to increase timeout for all tests at once?

IlyaSkriblovsky:
But there is no fails due to unfinished name resolution anymore :-)

IlyaSkriblovsky:
@psi29a can you restart tests manually a couple of times to prove there will be no random fails anymore?

psi29a:
I'm merging this because it works, apparently it failed last because of my pull request I merged in.

IlyaSkriblovsky:
Yeah, it seems like something messed with multiple versions of source code

PR #95, opened 2015-05-15
switch away from localhost

psi29a:
let's what happens :)


psi29a:
With the combination of my coverage and @IlyaSkriblovsky 's fixes, this should be green...

psi29a:
done

PR #97, opened 2015-05-21
Failover fixed and covered with unittests

IlyaSkriblovsky:
I've noticed that txmongo doesn't work with failover of replicaset as expected. It should raise AutoReconnect and client code should be able to re-issue query immediately.
Instead, txmongo raises ConnectionDone and immediate retry raises may fail with same exception.

Also, txmongo duplicated records in hosts list if same hosts were specified in connection URI and auto-discovered from replicaset config.

This PR fixes this and also introduces some testcases for replicaset.

The downside is that tests from `test_replicaset.py` are extremely slow. They are taking around 1 minute on my machine (combined) while all other tests take 3 seconds. That's because they need to deploy 3 mongod processes, configure replicaset, wait until it synced, and the slowest: wait until mongods will elect new primary in case of failover. TestCase.setUpClass()/tearDownClass() may speed things up, but they are not supported by `trial`.

### DIFF
--- a/txmongo/connection.py
+++ b/txmongo/connection.py
@@ -4,6 +4,7 @@
 
 from pymongo.errors import AutoReconnect, ConfigurationError, OperationFailure
 from pymongo.uri_parser import parse_uri
+from pymongo.read_preferences import ReadPreference
 
 from twisted.internet import defer, reactor, task
 from twisted.internet.protocol import ReconnectingClientFactory
@@ -15,7 +16,7 @@ from txmongo.protocol import MongoProtocol, Query
 
 class _Connection(ReconnectingClientFactory):
     __notify_ready = None
-    __discovered = None
+    __allnodes = None
     __index = -1
     __uri = None
     __conf_loop = None
@@ -26,7 +27,7 @@ class _Connection(ReconnectingClientFactory):
     maxDelay = 60
 
     def __init__(self, pool, uri, id):
-        self.__discovered = []
+        self.__allnodes = list(uri["nodelist"])
         self.__notify_ready = []
         self.__pool = pool
         self.__uri = uri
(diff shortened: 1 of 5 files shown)


txmongo/connection.py
```diff
@@ -39,18 +40,30 @@ def __init__(self, pool, uri, id):
...
+    def _initializeProto(self, proto):
+        yield proto.connectionReady()
+        self.resetDelay()
 
-        if not self.uri['options'].get('slaveok', False):
-            # Update our server configuration. This may disconnect if the node
-            # is not a master.
-            ready_deferred.addCallback(lambda _: self.configure(p))
+        uri_options = self.uri['options']
+        slaveok = uri_options.get('slaveok', False)
```
IlyaSkriblovsky: `readpreference` is contemporary replacement for `slaveok`

IlyaSkriblovsky:
Ouch, sounds like incompatibility of `replSetGetStatus` between MongoDB 2.x vs 3.0. Will look into it.

psi29a:
You saw the same thing I was during my tests in my branch here:
https://github.com/psi29a/txmongo/pull/5

I'm creating some tests to cover replication sets.

psi29a:
Hah, you added the same variable option in mongod ;)

psi29a:
In the interest of reducing the duplication of effort, maybe we should coordinate better. We've done a lot of the same work it seems.

psi29a:
"TestCase.setUpClass()/tearDownClass() may speed things up, but they are not supported by trial."

^-- This was removed from trial several years ago, we should just deal with the fact that it is never coming back. 
https://twistedmatrix.com/trac/ticket/4175

IlyaSkriblovsky:
Yeah, sorry for duplicating your efforts. I will ping you next time I'm going to work on something.

It's a pity that setUpClass was removed, it might be useful in testcases that require starting distinct mongod processes.

IlyaSkriblovsky:
The hard part of this PR is the fix that makes txmongo work well with AutoReconnect pattern:
```python
while True:
    try:
        conn.query()
        break
    except AutoReconnect:
        pass
```

Please check that you don't mind against changes to `txmongo/protocol.py`

psi29a:
Nice, I'll let travis do it's thing and merge when ready. I want to see the three-replicating mongos survive a second time all the way through.

IlyaSkriblovsky:
Ouch, seems that my code checking that replicaset is ready is not bulletproof

psi29a:
Great work!

PR #99, opened 2015-05-23
db.command() method

IlyaSkriblovsky:
Signature and behavior are same as in PyMongo

Almost all occurences of `db["$cmd"].find_one()` replaced with this new method.

Since `db.command()` has integrated simple error-checking, redundant checks were removed in some methods that are using this new method.

### DIFF
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -62,17 +62,17 @@ class TestMongoAuth(unittest.TestCase):
         conn = self.__get_connection()
 
         try:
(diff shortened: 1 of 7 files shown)


psi29a:
Nice work :) I'll be away tomorrow but I should be able to review it Monday.

I'm currently working on coverage for gridfs... finding weird things along the way. I'm also dropping for support for python <= 2.5 since it is no longer maintained... by anyone.

psi29a:
Looks good to me, I didn't see anything obvious. Again, great work!

PR #100, opened 2015-05-24
coll.rename() fix and additional tests

IlyaSkriblovsky:
I've added tests for `coll.rename()` and `coll.options()`

While testing `Collection.rename()` I've also noticed a typo in `Database.__call__()` that made `Collection.rename()` completely non-working :)


psi29a:
Nice find! :)

PR #101, opened 2015-05-25
Proper failure of create_collection instead of hang in case of error

IlyaSkriblovsky:
If `create_collection` is failing, it doesn't calls errback on deferred it returns, so calling code will hang forever


psi29a:
Nice catch there.

PR #102, opened 2015-05-25
Additional tests for collection.py and small fix for map_reduce()

IlyaSkriblovsky:
Before refactoring some methods of `collection.py` to use `inlineCallbacks`, i'd like to have each code path covered by tests.

Now `collection.py` is 100%-covered except for lines that should never  be reached or should not be reached with library versions we use in testing environment (pymongo==3.0).

While doing tests for `map_reduce()` I've found typo that made it useless without `full_response=True` :)

### DIFF
--- a/tests/test_collection.py
+++ b/tests/test_collection.py
@@ -21,7 +21,6 @@ from pymongo import errors
 
 from twisted.internet import defer
 from twisted.trial import unittest
-from unittest import SkipTest
 
 import txmongo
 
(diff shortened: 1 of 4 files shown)


psi29a:
It would seem that SkipTest doesn't work on some versions...

IlyaSkriblovsky:
Ahhh, 5 minutes

IlyaSkriblovsky:
Seems like timeout issue... Could you please restart failing configuration?

psi29a:
Nice work! :)

PR #105, opened 2015-05-26
inlineCallbacks everywhere

IlyaSkriblovsky:
I've refactored a bunch of methods to use inlineCallbacks-style where it is applicable.
Main exception is the `find_with_cursor()` which has unusual recurring deferreds semantics.

As a consequence, a couple of tests were altered to use `assertFailure` instead of `assertRaises` since `@inlineCallbacks` returns failed deferred in case of errros instead of direct raising exceptions.


IlyaSkriblovsky:
Coverage loss was expected just because total number of line decreased while number of missed lines remained the same

psi29a:
I reasoned as much, it happened with the previous commit and also with mine.

PR #106, opened 2015-05-27
find_one() modified to return None instead of {} when nothing found

IlyaSkriblovsky:
This change relates only to `find_one`, `find` still returns empty list when nothing was found


IlyaSkriblovsky:
Looks like we have some lines which have unstable coverage. Sometimes they are visited during some tests and sometimes not. I mean these: https://coveralls.io/builds/2651523/source?filename=txmongo%2Fprotocol.py#L322
They were occasionally visited during previous build but not visited now, so coverage is decreased a bit.

psi29a:
I've seen those as well, likely the way those versions for Twisted work.

PR #108, opened 2015-05-29
Recently added features fixed to be compatible with PyMongo 2.8

IlyaSkriblovsky:
A couple of new features appear to be broken on PyMongo 2.8. This patch fixes compatibility.


psi29a:
I'll let travis do his thing, if all is OK I'll merge it in. Nice catch. No worries about coverage. ;)

IlyaSkriblovsky:
Some race condition in replicaset tests again...

PR #109, opened 2015-05-30
Write concern

IlyaSkriblovsky:
API closely resembles PyMongo

* `MongoConnection` and `ConnectionPool` now takes write-concern options as named args and as URI options (named args take precedence)

* You can set override write-concern for `Database` and `Collection` when creating them using constructors (`db = Database(conn, "dbname", write_concern=WriteConcern(...)`). In this case write concern is specified using instance of special `WriteConcern` class. It is copied to txmongo source tree from PyMongo 3.0 because it doesn't exist in 2.8.

* `insert`, `update`, `save` and `remove` now takes both `safe` argument and write-concern named arguments. Semantics is the same as in PyMongo 2.x:
    * named args (`w`, `wtimeout`, `j`, `fsync`) take precedence over `safe`
    * `safe=False` forces query to be unacknowledged
    * `safe=True` uses inherited (connection-, db- or collection-level) write-concern if it's acknowledged or forces simple acknowledged write-concern otherwise
    * if no arguments if specified inherited write-concern is used

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -10,3 +10,4 @@ wheel
 setuptools
 tox
 check-manifest
+mock
--- a/tests/test_write_concern.py
+++ b/tests/test_write_concern.py
@@ -1,0 +1,152 @@
+# coding: utf-8
+# Copyright 2015 Ilya Skriblovsky <ilyaskriblovsky@gmail.com>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from mock import Mock, patch
+from twisted.internet import defer
+from twisted.trial import unittest
+from txmongo.connection import MongoConnection, ConnectionPool
+from txmongo.protocol import MongoProtocol
+from txmongo.database import Database
+from txmongo.collection import Collection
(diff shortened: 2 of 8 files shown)


IlyaSkriblovsky:
I've also created PR #110 that might fix spontaneous replicaset test failures like this one

psi29a:
I merged the two other PRs and rerun travis on this PR... we got at least 2 failures, but they are both attributed to exit 48, socket problem.

IlyaSkriblovsky:
May be I need to rebase this branch to get benefit from #110

psi29a:
That makes this PR go through the travis cycle 3 times now. :)

Looks good to me.

psi29a:
Can you also update the NEW.rst with new features and bug fixes please? :)

IlyaSkriblovsky:
Yep: #112

IlyaSkriblovsky:
Ouch, -2.5% of coverage? That's strange...

psi29a:
Looks like mostly type checking code that isn't covered:
https://coveralls.io/builds/2692471/source?filename=txmongo%2Fwrite_concern.py

anyway to try to get through this "unhappy" path? ;)

IlyaSkriblovsky:
Ok, I will add tests for that. But when we eventually drop support for PyMongo 2.8 we will use original `pymongo.write_concern.WriteConcern` class instead, so both `write_concern.py` and corresponding tests will be removed.

psi29a:
When do you think that'll happen? How long do you think we should keep 2.8 compatibility?

IlyaSkriblovsky:
I'm working now on new *_one/*_many query methods which returns various `InsertOneResult`, `InsertManyResult`, `UpdateResult` objects. Currently, I'm importing them from pymongo, but they exist only in pymongo 3.0.
So, when merging this code we will need to choose: 1) drop support for PyMongo 2.x; or 2) copy these classes to TxMongo source code like I've copied `WriteConcern` class

IlyaSkriblovsky:
New coverage here: #113

PR #110, opened 2015-05-30
Trying to improve durability of test_replicaset.py

IlyaSkriblovsky:
Sometimes replicaset tests are failing with
> namespace admin.$cmd failed: all members and seeds must be reachable to initiate set

This is probably because `setUp()` method does `rs.initiate()` right after several mongod processes are started. This patch tries to address this by connecting to each started mongod to ensure all replicaset members are up and running before running `rs.initiate()`.

### DIFF
--- a/tests/test_replicaset.py
+++ b/tests/test_replicaset.py
@@ -46,9 +46,18 @@ class TestReplicaSet(unittest.TestCase):
         return d
 
     @defer.inlineCallbacks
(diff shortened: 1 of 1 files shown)


IlyaSkriblovsky:
May be it worth to try to retry tests several times to check if this exception is gone

psi29a:
restarted, if nothing came up, I'll merge it in.

PR #115, opened 2015-06-02
New query methods

IlyaSkriblovsky:
Implementation of new-style query methods of PyMongo 3.0:
- `insert_one`, `insert_many`
- `update_one`, `update_many`
- `replace_one`, `replace_many`
- `delete_one`, `delete_many`
- `find_one_and_delete`, `find_one_and_update`, `find_one_and_replace`

All new methods (except `find_one_and_***` ones) return `pymongo.results.***Result` instances just like PyMongo's counterparts.

**WARNING: This code breaks compatibility with PyMongo 2.x** by using classes which are only available in PyMongo 3.0

List of what we use from PyMongo 3.0:
- `pymongo.results.***Result`
- `pymongo.common.validate_ok_for_update`, `pymongo.common.validate_ok_for_replace`
- `pymongo.collection.ReturnDocument`

All these are just simple helper classes and functions. So we have two options:

1. Drop support with PyMongo 2.x. The downside is that PyMongo 3.0 have several backward-incompatible API changes, for example, `safe=` kwarg was removed from good old CRUD methods. So if one is using both PyMongo and TxMongo in same virtualenv, she will need to modify theirs existing synchrous code which uses PyMongo.

2. Keep support with PyMongo 2.x and just copy-paste listed stuff from PyMongo 3.0 into TxMongo's codebase. Not so pretty though, but works and reduces efforts needed from users to upgrade. (Btw, we already did this way for `txmongo.write_concern.WriteConcern`).

What do you think?

*Note: breaking compatibility with PyMongo 2.x have nothing to do with breaking compabitibility with MongoDB 2.x. MongoDB 2.6 is still supported by PyMongo 3.0*

### DIFF
--- a/tests/test_queries.py
+++ b/tests/test_queries.py
@@ -15,11 +15,13 @@
 
 from bson import BSON, ObjectId
 from bson.son import SON
-from pymongo.errors import OperationFailure
+from pymongo.errors import OperationFailure, WriteError
+from pymongo.results import InsertOneResult, InsertManyResult, UpdateResult
 from twisted.internet import defer
 from twisted.trial import unittest
 import txmongo
 from txmongo.protocol import MongoClientProtocol
+from txmongo.write_concern import WriteConcern
 
 mongo_host = "localhost"
 mongo_port = 27017
@@ -35,15 +37,22 @@ class _CallCounter(object):
         return self.original(this, *args, **kwargs)
 
 
-class TestMongoQueries(unittest.TestCase):
+class _SingleCollectionTest(unittest.TestCase):
 
-    timeout = 15
-
-    @defer.inlineCallbacks
     def setUp(self):
-        self.conn = yield txmongo.MongoConnection(mongo_host, mongo_port)
+        self.conn = txmongo.MongoConnection(mongo_host, mongo_port)
         self.db = self.conn.mydb
         self.coll = self.db.mycol
+
+    @defer.inlineCallbacks
+    def tearDown(self):
+        yield self.coll.drop()
+        yield self.conn.disconnect()
+
+
+class TestMongoQueries(_SingleCollectionTest):
+
+    timeout = 15
 
     @defer.inlineCallbacks
     def test_SingleCursorIteration(self):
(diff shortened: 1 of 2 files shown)


tests/test_queries.py
```diff
@@ -622,3 +549,423 @@ def test_MapReduce(self):
         self.assertTrue(result["ok"], 1)
         self.assertTrue("counts" in result)
         self.assertTrue("results" in result)
+
+
+class TestInsertOne(_SingleCollectionTest):
+
+    @defer.inlineCallbacks
+    def test_Acknowledged(self):
```
IlyaSkriblovsky: New `***_one/many` methods are using completely different protocol depending on whether current write-concern is acknowledged or not. If it is acknowledged, they are using new db commands which were introduced in MongoDB 2.6. When unacknowledged, they are using good old OP_INSERT/OP_UPDATE/... messages. So for each query methods we are testing ack'ed and unack'ed separately because message format and also additional flags handling (upsert, ordered, ...) are completely different.

txmongo/collection.py
```diff
@@ -479,3 +611,50 @@ def find_and_modify(self, query=None, update=None, upsert=False, **kwargs):
...
+        if upsert is not None:
+            validate_boolean("upsert", upsert)
+            cmd["upsert"] = upsert
+
+        no_obj_error = "No matching object found"
+
+        result = yield self._database.command(cmd, allowable_errors=[no_obj_error])
+        defer.returnValue(result.get("value"))
+
+    def find_one_and_delete(self, filter, projection=None, sort=None, **kwargs):
```
IlyaSkriblovsky: There is a terminology clash here. TxMongo traditionally used `filter` kwarg for specifying sorting, hinting and other modifiers. But all new methods use `filter` argument to specify query and `sort` argument to specify sorting (if applicable). So we need to choose whether to be consistend with the rest of txmongo's method args or to be consistend with PyMongo. For now I choosed second option and named all args just like in PyMongo, but it might be changed if there will be objections.

trenton42: Personally, I would rather stay closer to pymongo. I occasionally have to switch to synchronous methods (in ipython, for example), and it's nice to have them be somewhat close. It also feels closer to the mongo shell.

psi29a:
@fiorix adding to conversation, what do you think?

psi29a:
@trenton42 also, your opinion here is important :)

txmongo/collection.py
psi29a: I suggest that we stay so close as possible to PyMongo.  I hate to conflict with their naming convention and we should adjust ours to make it easier for those wanting to port their existing code or write new code.

psi29a:
I'm honestly of the opinion of following PyMongo.  People can always use the older version of TxMongo (15.0) and then update their code when ready to the latest standard. They'll be doing that with PyMongo as well.

IlyaSkriblovsky:
So it seems to be consensus here :+1: 

But what about breaking/not breaking compatibility with PyMongo 2.x?

psi29a:
I'm OK with breaking <= 2.8 compatibility.  Don't forget that 2.9 will be released to bridge 3.0 and 2.8 code, back-porting many things from 3.0

Maybe this is something to investigate?

IlyaSkriblovsky:
2.9 doesn't have any of listed features as I can see from `v2.9` branch on theirs repo :(
Seems that they are not going to backport new-style query methods, so all these helper classes and functions are not needed in 2.9

fiorix:
Agreed

psi29a:
Right, moving forward. Accepting this PR also means that for the *_one and *_many we're moving towards PyMongo 3.0 style.

PR #117, opened 2015-06-04
Drop pymongo 2.x compatibility hacks

IlyaSkriblovsky:
As we drop pymongo 2.x support anyway, lets get rid of hacky workarounds and 2.x/3.x detection code

I also removed unittest which was solely testing ``txmongo.write_concern.WriteConcern`` class which is now replaced by PyMongo's one.


psi29a:
Not worried about coverage decrease, we're using less code. :)

PR #118, opened 2015-06-05
Testing MONGODB-CR auth method

IlyaSkriblovsky:
`test_auth.py` expoits a hack to force MongoDB to use obsolete MONGODB-CR auth mechanism in order to get coverage of both SCRAM-SHA-1 and MONGODB-CR auth methods.

Travis environment changed to use MongoDB 3.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ matrix:
 
 before_install:
   - "sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 7F0CEB10"
-  - "echo 'deb http://downloads-distro.mongodb.org/repo/ubuntu-upstart dist 10gen' | sudo tee /etc/apt/sources.list.d/mongodb.list"
+  - "echo 'deb http://repo.mongodb.org/apt/ubuntu '$(lsb_release -sc)'/mongodb-org/3.0 multiverse' | sudo tee /etc/apt/sources.list.d/mongodb-org-3.0.list"
   - "sudo apt-get update"
   - "sudo apt-get install mongodb-org-server"
   - "mongod --version"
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -45,7 +45,7 @@ matrix:
 
 before_install:
   - "sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 7F0CEB10"
-  - "echo 'deb http://downloads-distro.mongodb.org/repo/ubuntu-upstart dist 10gen' | sudo tee /etc/apt/sources.list.d/mongodb.list"
+  - "echo 'deb http://repo.mongodb.org/apt/ubuntu '$(lsb_release -sc)'/mongodb-org/3.0 multiverse' | sudo tee /etc/apt/sources.list.d/mongodb-org-3.0.list"
```
psi29a: Nicely found!

psi29a:
95% that's really great!

psi29a:
I'm only worried about that we're no longer testing mongodb >= 3.0, so the noDups test will always be skipped.

IlyaSkriblovsky:
> we're no longer testing mongodb >= 3.0

\< 3.0 you meant?
Is it possible to specify different configurations for Travis? I think it won't be easy to install both versions of MongoDB side-by-side...

psi29a:
Yeah, sorry. That is what I meant. I thought it could be possible to install both. If not, I could create a PPA package, statically compiled, so that it doesn't interfere with the other installed MongoDB.

Would that work, or overkill?

IlyaSkriblovsky:
Actually, I think it is overkill because it's all about testing two pretty obvious lines of code :)

psi29a:
;)

Well, as an option, there is always this:
curl -O http://downloads.mongodb.org/linux/mongodb-linux-x86_64-2.6.10.tgz
^-- statically compiled

PR #120, opened 2015-06-06
Test handling invalid types of username/password

IlyaSkriblovsky:
To make 100% coverage of `database.py` :)


IlyaSkriblovsky:
Seems like I need to investigate that "return code 48" fails of test_replica set.py...

PR #121, opened 2015-06-06
Some unreachable code removed

IlyaSkriblovsky:


txmongo/connection.py
```diff
@@ -157,8 +157,6 @@ def notifyReady(self):
             """
         if self.instance:
             return defer.succeed(self.instance)
-        if self.__notify_ready is None:
-            self.__notify_ready = []
```
IlyaSkriblovsky: `__notify_ready` can never be None since it's only set to lists

txmongo/database.py
```diff
@@ -7,15 +7,6 @@
...
 
-# PyMongo 2.x and 3.x has a bit different signatures of _check_command_response
-# which differ only in `reset` argument that is not used by txmongo
-import inspect
-if 'reset' in inspect.getargspec(_check_command_response).args:
-    orig = _check_command_response
-    def wrapper(response, msg=None, allowable_errors=None):
-        return orig(response, None, msg=msg, allowable_errors=allowable_errors)
-    _check_command_response = wrapper
-
```
IlyaSkriblovsky: This is not needed since we dropped support for PyMongo 2.x

psi29a:
Good catches.

PR #122, opened 2015-06-09
Collection props

IlyaSkriblovsky:
`name`, `full_name` and `database` properties added to Collection class along with unittests for them

The meanings are exactly the same as in PyMongo


psi29a:
Great work again! :)

PR #123, opened 2015-06-09
Remove dbref

IlyaSkriblovsky:
Is there any reason for TxMongo to contain it's own `DBRef` implementation?

It was probably copied-and-pasted from older version of PyMongo. I suggest to remove it and use PyMongo's `bson.dbref.DBRef` instead.

But the is a couple of incompatibilities here:
- TxMongo's `DBRef` can accept `Collection` instance as its first argument. PyMongo's `DBRef` only accept strings;
- TxMongo's `DBRef` declared `__cmp__` for some mysterious reason. PyMongo's `DBRef` cannot be meaningfully compared.

I think both these issues are trivial enough to be neglected in favor of more code sharing with PyMongo.

What do you think?

p.s. coverage decrease is expected :)


fiorix:
I wish I could remember why. There's probably something on Git logs.

IlyaSkriblovsky:
It seems that TxMongo originally used pymongo's `DBRef`, but then in d769c58e0ccfc723a1872d111e1fb17cb1614779 it was copied it to its own source tree and altered to add ability to pass `Collection` instance in `__init__` instead of collection name string.

PyMongo's `DBRef` doesn't allow that. I agree that it's a cool feature, but do we need it with the price of incompatibility and code duplication?

IlyaSkriblovsky:
FYI:  `__cmp__` was replaced by `__eq__` in `bson.dbref` between 2.1.1 and 2.2
https://github.com/mongodb/mongo-python-driver/commit/7474f5cde80fb0bd0e4a1bf4cbb88b4902131810

psi29a:
Honestly, we should be following PyMongo's lead here. We don't have the resources of the official client. I'm OK with this.

PR #124, opened 2015-08-07
Port txmongo to Python 3.3+

hawkowl:


fiorix:
Rebase it maybe? You did some stuff in c7f430c then changed back in b813711, etc.

IlyaSkriblovsky:
Nice job!

Seems like something wrong with testing environment configuration. Travis is trying to test py3-* configurations with python 2.7: https://travis-ci.org/twisted/txmongo/jobs/74519700#L92
```
$ source ~/virtualenv/python2.7/bin/activate
$ python --version
Python 2.7.9
$ pip --version
pip 6.0.7 from /home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages (python 2.7)
```

Though, I'm not sure it is possible at all to run our unittests with Py3 because Trial is not ported to Py3 yet. I thought about porting TxMongo to py3 before, but this stopped me from trying.

Or I'm wrong about Trial? I would be happy to.

IlyaSkriblovsky:
Nice job!

Wow, I've missed a moment when Trial was ported ty Py3

hawkowl:
Trial is ported in trunk -- 15.4 will be the first release, which is why the travis only works on trunk.

Fixing up the auth tests now...

IlyaSkriblovsky:
Seems like all tests which talks to DB are hanging

psi29a:
Thanks for this! :)  Seems to be hanging on python33 and python34 though,  We'll have to dig a bit deeper on that.

hawkowl:
I have "PASSED (skips=1, successes=144)" on local 3.3 and 3.4. The authentication tests were busted which I think caused a knock-on. Lets see how these go.

IlyaSkriblovsky:
That's strange, but when I run tests with Py3 on vagrant machine, all tests which talk to DB are hanging. But if I change "localhost" to "127.0.0.1" in test files, they work well (except auth tests which is distinct issue). Seems like same thing breaks tests on Travis.

hawkowl:
Try updating, the auth tests work now with the latest branch.

I'm about to jet back to Perth, and this patch probably needs a bit more work -- making all trunk builds on allowed failures, moving pyflakes off allowed failures, and some code cleanup + fixing that minor coverage regression.

tests/test_queries.py
```diff
@@ -21,11 +23,15 @@
...
 from pymongo.write_concern import WriteConcern
 from twisted.internet import defer
+from twisted.python.compat import _PY3
 from twisted.trial import unittest
 import txmongo
 from txmongo.protocol import MongoClientProtocol
 import txmongo.filter as qf
 
+if _PY3:
+    from twisted.python.compat import xrange
```
IlyaSkriblovsky: It would be cleaner to just use `range` instead of `xrange`

tests/test_queries.py
```diff
@@ -170,7 +176,7 @@ def __check_no_open_cursors(self):
     @defer.inlineCallbacks
     def test_CursorClosing(self):
         # Calculate number of objects in 4mb batch
-        obj_count_4mb = 4 * 1024**2 / len(BSON.encode(self.__make_big_object())) + 1
+        obj_count_4mb = int(4 * 1024**2 / len(BSON.encode(self.__make_big_object())) + 1)
```
IlyaSkriblovsky: Better use `//` instead of `int(... / ...)` here

IlyaSkriblovsky:
Seems like name resolution is broken in Twisted trunk. Following code works on Twisted 15.3.0, but fails on trunk:
```
def done(result):
    print('done', result)
    reactor.stop()

reactor.resolve('localhost').addCallback(done)
reactor.run()                                                                                                                                                                                                        
```

```
  File "/vagrant/testresolve.py", line 7, in <module>
    reactor.resolve('localhost').addCallback(done)
  File "/home/vagrant/pyenv3/lib/python3.4/site-packages/twisted/internet/base.py", line 570, in resolve
    return self.resolver.getHostByName(name, timeout)
  File "/home/vagrant/pyenv3/lib/python3.4/site-packages/twisted/internet/base.py", line 271, in getHostByName
    self.reactor, self.reactor.getThreadPool(),
  File "/home/vagrant/pyenv3/lib/python3.4/site-packages/twisted/internet/base.py", line 991, in getThreadPool
    self._initThreadPool()
  File "/home/vagrant/pyenv3/lib/python3.4/site-packages/twisted/internet/base.py", line 956, in _initThreadPool
    from twisted.python import threadpool
  File "/home/vagrant/pyenv3/lib/python3.4/site-packages/twisted/python/threadpool.py", line 16, in <module>
    from twisted._threads import pool as _pool
  File "/home/vagrant/pyenv3/lib/python3.4/site-packages/twisted/_threads/__init__.py", line 15, in <module>
    from ._pool import pool
ImportError: No module named 'twisted._threads._pool'
```

That's why all our tests that try to connect to MongoDB on `localhost` are hanging.

We can work around this by changing "localhost" to "127.0.0.1", but I'd prefer to wait when Twisted trunk will be fixed.

txmongo/collection.py
```diff
@@ -2,7 +2,7 @@
 # Use of this source code is governed by the Apache License that can be
 # found in the LICENSE file.
 
-import types
+from __future__ import absolute_import, division
```
IlyaSkriblovsky: We have no division operations in this file :)

hawkowl: I'm going off the Twisted py3 porting standards (https://twistedmatrix.com/trac/wiki/Plan/Python3#Reviewerchecklist) where it's in there regardless. If that's an issue, I can change it.

hawkowl:
The tests pass! Just have to figure out what coveralls is complaining about...

txmongo/collection.py
IlyaSkriblovsky: Ok, I don't mind at all. Just thought this is a trace of some changes that were reverted.

psi29a:
That coveralls decreased by 0.02% is no big deal, we're still above 95%.

Hawky, you feel comfortable with merging it anyway? :)

hawkowl:
I think it's in a merging state, if you do. :)

IlyaSkriblovsky:
Coveralls is not an issue: we have 3 lines in `protocol.py` which can be passed or missed with some probability during replicaset tests.

@hawkowl, I saw you have fixed Twisted trunk to make (in particular) DNS resolution work. Thanks!

psi29a:
Good to go for us, thank you very much @hawkowl :D

hawkowl:
No problem :)

PR #128, opened 2015-09-15
Unicode fix

psi29a:


tests/test_connection.py
```diff
@@ -44,3 +43,14 @@ def test_Misc(self):
...
         self.assertTrue(len(result[0].uri['nodelist']) == 0)
         self.assertEqual("Connection()", repr(self.named_conn))
+
+    @defer.inlineCallbacks
+    def test_uri_input(self):
+        test = txmongo.connection.ConnectionPool()
+        yield test.disconnect()
+        test = txmongo.connection.ConnectionPool("mongodb://127.0.0.1/dbname")
+        yield test.disconnect()
+        test = txmongo.connection.ConnectionPool(u"mongodb://127.0.0.1/dbname")
```
IlyaSkriblovsky: Under Python 3 these two checks will be the same. May be use `b"..."` for the first one to test byte-string too? Or we are not going to support bytes in Py3?

txmongo/connection.py
```diff
@@ -241,7 +241,7 @@ class ConnectionPool(object):
     __wc_possible_options = set(['w', "wtimeout", 'j', "fsync"])
 
     def __init__(self, uri="mongodb://127.0.0.1:27017", pool_size=1, ssl_context_factory=None, **kwargs):
-        assert isinstance(uri, str)
+        assert isinstance(uri, StringType)
```
IlyaSkriblovsky: There is a convention to raise `TypeError` when arguments of invalid types are passed... But I agree that `assert` looks shorter and cleaner. Anyways, it is out of scope of this pull.

txmongo/gridfs.py
```diff
@@ -3,7 +3,6 @@
 # found in the LICENSE file.
 
 from __future__ import absolute_import, division
-
 from txmongo._gridfs import GridFS, GridIn, GridOut, errors, grid_file, GridOutIterator
 
 assert GridFS
```
IlyaSkriblovsky: What are we checking here? :)

psi29a: Heh... the asserts are for silly coverage reasons but aren't necessary. NOOPs.

I honestly want to either get rid of gridfs or if it is used by anyone (hello?!) refactored.

txmongo/connection.py
psi29a: I would rather get rid of all asserts and just raise errors. We can do that in another refactor branch?

IlyaSkriblovsky: Sure, I've just noticed while reading diff. Don't need to fix it in this branch.

tests/test_connection.py
psi29a: That's OK, we are trying to find problems between python2 and python3 with regard to the default string behaviour. 

Adding a b"blah" in python2 will work just fine, but in python3 it will error out an AssertionError.

psi29a: Was just reading:
The new bytes type is 3.x only. The 2.x bytes built-in is just an alias to the str type. There is no new type called bytes in 2.x; Just a new alias and literal syntax for str.

So bytes, was never really used in py2, it just remapped to str.

PR #130, opened 2015-09-18
raise NotMasterError to indicate to client that it is safe to retry, …

psi29a:
…cleanup, add explicit test for AutoReconnect

### DIFF
--- a/tests/test_replicaset.py
+++ b/tests/test_replicaset.py
@@ -15,17 +15,14 @@
 
 from __future__ import absolute_import, division
 
+from bson import SON
 from pymongo.errors import OperationFailure, AutoReconnect, ConfigurationError
 from twisted.trial import unittest
-from twisted.python.compat import _PY3
 from twisted.internet import defer, reactor
 from txmongo.connection import MongoConnection, ConnectionPool, _Connection
 from txmongo.protocol import QUERY_SLAVE_OK, MongoProtocol
 
 from .mongod import Mongod
-
-if _PY3:
-    from twisted.python.compat import xrange
 
 
 class TestReplicaSet(unittest.TestCase):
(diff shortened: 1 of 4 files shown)


txmongo/collection.py
```diff
@@ -297,12 +297,11 @@ def _get_write_concern(self, safe=None, **wc_options):
                 # In this case safe=True must issue getLastError without args
                 # even if connection-level write concern was unacknowledged
                 return WriteConcern()
-        elif safe == False:
+        elif not safe:
```
IlyaSkriblovsky: may be just `else` here? and remove final `return self.write_concern`. It is unreachable with this `elif not safe` anyways

txmongo/collection.py
```diff
@@ -394,7 +392,6 @@ def update(self, spec, document, upsert=False, multi=False, safe=None, flags=0,
             ret = yield proto.get_last_error(str(self._database), **write_concern.document)
             defer.returnValue(ret)
 
-
```
IlyaSkriblovsky: The main think I hate about PEP8 is that it prohibits double empty lines between methods :(

tests/test_replicaset.py
```diff
@@ -144,23 +143,38 @@ def test_AutoReconnect(self):
...
+        if _version.version.major < 15:
+            raise unittest.SkipTest("No 'with self.assertRaises' in Twisted below version 15.")
+        self.patch(_Connection, 'maxDelay', 5)
+        uri = "mongodb://localhost:{0}/?w={1}".format(self.ports[0], len(self.ports))
+        conn = ConnectionPool(uri)
+
+        # this will force primary to step down, triggering an AutoReconnect that bubbles up
+        # through the connection pool to the client
+        with self.assertRaises(AutoReconnect):
+            yield conn.admin.command(SON([('replSetStepDown', 86400), ('force', 1)]))
```
IlyaSkriblovsky: You may use `assertFailure` here

IlyaSkriblovsky: And sorry, I suggested you an ugly form of `command()`. It can be written nicier: `command("replSetStepDown", 86400, force=1)`

psi29a: assertFailure on a yield?

IlyaSkriblovsky: Like this: `yield self.assertFailure(conn.admin.command(...), AutoReconnect)`

PR #132, opened 2015-09-28
Better handling of deadline and timeout

psi29a:


tests/test_queries.py
```diff
@@ -290,13 +293,13 @@ def test_LimitAtBatchThresholdEdge(self):
         res = yield self.coll.find(limit=100)
         self.assertEqual(len(res), 100)
 
-        yield self.coll.drop(safe=True)
+        yield self.coll.drop()
```
psi29a: In case anyone is wondering, pymongo never supported safe in drop and drop_collection... neither did we, it was just sucked up into **kwargs and ignored later on.
http://api.mongodb.org/python/current/api/pymongo/collection.html#pymongo.collection.Collection.drop

txmongo/database.py
```diff
@@ -44,14 +45,17 @@ def connection(self):
...
         if isinstance(command, (bytes, unicode)):
             command = SON([(command, value)])
-        command.update(kwargs)
+        options = kwargs
+        options.pop("_deadline", None)
+        command.update(options)
 
         ns = self["$cmd"]
-        response = yield ns.find_one(command)
+        response = yield ns.find_one(command, **kwargs)
```
IlyaSkriblovsky: `_deadline` is not passed to `find_one`, because you popped it from `kwargs`. Did you meant this?

IlyaSkriblovsky:
< grumbling >
Actually, I don't much like idea of adding some sort of hidden arguments to each query method.

This way it would be hard to write new method that would receive `**kwargs` and do something with it as with `dict` (like `command()` or `distinct()` create DB query from `kwargs`) without hidden knowledge that some sort of underscored arguments may appear in `kwargs` and needs to be removed aforehead. 

Or if someone will introduce new query method and will fail to follow guideline of receiving `**kwargs` and mandatory passing it to all subcalls, the deadline feature will be broken. And this guideline isn't obvious from the code without studying internals of `@timeout`. This looks like a sign of fragile design that is easy to break.

You changes for supporting deadline is pretty broad already. Isn't it better to remove `inlineCallbacks` from query methods and get benefit from Twisted's builtin cancellation mechanism? This would be a massive change too, but it will allow us to keep deadline feature separate from internals of _each_ query method.
</ grumbling >

sorry :)

txmongo/database.py
psi29a: You're right, it needs to be put back in.  I hate this sneaky shit.

PR #138, opened 2015-11-13
?authMechanism URI option support backported from 15.1.0 to 15.0.0

IlyaSkriblovsky:
I didn't cherry-picking because original tests for auth mechanisms in 15.1.0 require `db.command` which depend on PyMongo 3.0. So I added another testing code.

Is it ok that I inserted 15.0.1 in NEWS.rst?

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -1,6 +1,17 @@
 Changelog
 =========
 
+Release 15.0.1 (2015-11-13)
+---------------------------
+
+This is a maintenance update to outdated 15.0.0 release
+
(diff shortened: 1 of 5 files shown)


psi29a:
The errors look normal with pypy and py26-latest since python26 was dropped.

IlyaSkriblovsky:
So, ok to merge?

psi29a:
Looks OK, I'll merge and try to create a tag that we can then upload to pypi in a bit (today).

IlyaSkriblovsky:
Thanks! Let me know about success

PR #140, opened 2016-02-18
Circular reference removed between after_reply and its closures

IlyaSkriblovsky:
I was investigating how my server consumes RAM and noticed that each call to `find()` or `find_one()` creates unreachable reference cycle and this is (among with other reasons) why my app is constantly triggers GC.

`after_reply` inside `find_with_cursor` does some sort of deferred recursion and so references itself. This creates unobvious reference cycle between function object of `after_reply` and its closure.

Proposed solution is to break self-reference of `after_reply` by passing this reference into function via arguments. Rather tricky though, but less intrusive than moving `after_reference` to outer scope to remove closure.

Testing code:
```python
gc.disable()
print(len(gc.get_objects()))
for i in range(10000):
    yield coll.find_one()
print(len(gc.get_objects()))
```

Before:
```
19813
89630
```

After:
```
19796
19611
```

Since the default GC threshold is 700, "before" version triggers GC pretty often.

### DIFF
--- a/txmongo/collection.py
+++ b/txmongo/collection.py
@@ -203,10 +203,14 @@ class Collection(object):
                           query=spec, fields=fields)
 
             deferred_query = protocol.send_QUERY(query)
-            deferred_query.addCallback(after_reply, protocol)
+            deferred_query.addCallback(after_reply, protocol, after_reply)
             return deferred_query
 
-        def after_reply(reply, protocol, fetched=0):
+        # this_func argument is just a reference to after_reply function itself.
+        # after_reply can reference to itself directly but this will create a circular
+        # reference between closure and function object which will add unnecessary
+        # work for GC.
+        def after_reply(reply, protocol, this_func, fetched=0):
             documents = reply.documents
             docs_count = len(documents)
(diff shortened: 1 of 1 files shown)


psi29a:
Great find!

IlyaSkriblovsky:
Yes, in pythonic meaning of the term. GC will be able to clean this of course, but it's better to not summon it without a need. 10k of `find_one` was triggering GC ~80 times, with this fix it doesn't at all.

trenton42:
Wow, that is a great find! I have been puzzling over memory usage for quite a while, but I never got anywhere with it.

Any Idea when the next pypi release will be?

IlyaSkriblovsky:
@psi29a, do you think this fix deserves a minor release? I can fill the changelog if so

psi29a:
I believe it does, yes. Please do.  I'm in the middle of a release here at work so this is perfect timing.

We should make this 16.0, since it is 2016 now and well... memory 'leaks' are rather important. ;)

IlyaSkriblovsky:
Opened #141 for 16.0.0

Issue #142, opened 2016-03-01
Logging, we can do better.


psi29a:
In my logs, I get the following:
> 2016-03-01 16:53:43.398197 +0100 ERROR [] 'Connection lost.'
> 2016-03-01 16:53:43.399087 +0100 ERROR [] 'Connection lost.'

which come about when there aren't any mongodbs left alive:
https://github.com/twisted/txmongo/blob/master/txmongo/protocol.py#L325

Normally this would be fine, but for people reading logs (for the fun of it?), they won't have any context as to what is going on so we should provide better error messages such as:
'TxMongo lost connection to MongoDB.

It would probably not be so bad an idea to have something in connectionReady that say that TxMongo is ready as some TxMongo users (like me) put all all the twisted protocol log.msg to DEBUG to help cut down on log spam (3 messages per attempt to re-connect).

PR #143, opened 2016-03-02
better logging and version bump

psi29a:
This should help to address this: https://github.com/twisted/txmongo/issues/142

Feedback appreciated. :)


IlyaSkriblovsky:
Looks good. I also was many times guessing who is logging some strange errors into my log files before understanding that this was txmongo.

psi29a:
I'm glad I'm not the only one! ;)

Issue #145, opened 2016-04-05
insert_many with BulkWriteError


jhonny111s:
When use insert_may with ordered=False I expected have in BulkWriteError whith all duplicated keys but only return the last, how can i get all error (knowing data that induce error)

IlyaSkriblovsky:
Yes, I've checked this and it looks like txmongo raises invalid exceptions for `insert_many` and possibly other `*_many` methods (`DuplicateKeyError` instead of `BulkWriteError`).

We call `pymongo.helpers._check_write_command_response` that unwraps Mongo's error response into single `DuplicateKeyError`. This is correct for `*_one` methods, but not for `*_many`.

IlyaSkriblovsky:
Fixed this in #146. Could you please check if that branch raises expected exception type for you?

jhonny111s:
I use pip install txmongo but this install older master, how can install this branch. thanks

IlyaSkriblovsky:
Please try `pip install https://github.com/IlyaSkriblovsky/txmongo/archive/insert-many-bulk-errors.zip`

jhonny111s:
@IlyaSkriblovsky  thanks so much, works fine for me

IlyaSkriblovsky:
@jhonny111s This patch was merged into master, so you can use `pip install https://github.com/twisted/txmongo/archive/master.zip` until new release of txmongo will be published to PyPI

PR #146, opened 2016-04-06
Raising BulkWriteError from insert_many instead of WriteError/DuplicateKeyError

IlyaSkriblovsky:
As @jhonny111s noticed in #145, txmongo raises `WriteError`/`DuplicateKeyError` from `insert_many`, while PyMongo raises `BulkWriteError` which allows to check which documents are not inserted.

This patch fixes makes `insert_many` to always raise `BulkWriteError` like PyMongo does, and also adds additional tests for checking exception types from all `*_one`/`*_many` methods

### DIFF
--- a/tests/test_queries.py
+++ b/tests/test_queries.py
@@ -17,7 +17,8 @@ from __future__ import absolute_import, division
 
 from bson import BSON, ObjectId
 from bson.son import SON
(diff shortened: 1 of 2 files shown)


psi29a:
Looks great! Thank you!

PR #153, opened 2016-04-15
Insert more than 1000 documents with insert_many

### LINKED ISSUE #149: Txmongo is unable to handle more than 1000 objects in bulk operations
Hi. I'm testing the following code using txmongo 16.1.0. Basically, I'm sending 1001 objects to an insert_many.  It's raising pymongo.errors.OperationFailure: TxMongo: command SON ... ('writeConcern', {})]) on namespace foo.$cmd failed with 'exceeded maximum write batch size of 1000'
``` python
import time
import txmongo
from twisted.internet import defer, reactor


@defer.inlineCallbacks
def example():
    mongo = yield txmongo.connection.ConnectionPool('mongodb://127.0.0.1:27018')

    foo = mongo.foo  # `foo` database
    test = foo.test  # `test` collection

    # insert some data
    insert_data = []
    for x in range(1001):
        insert_data.append({"something": x*time.time()})

    result = yield test.insert_many(insert_data)

if __name__ == '__main__':
    example().addCallback(lambda ign: reactor.stop())
    reactor.run()
```
Mongo has a limit of 1000 for Bulk operations, but drivers usually handle this by dividing batch operations. Pymongo does not fail with many more objects.

IlyaSkriblovsky:
`insert_many` reimplemented to make it able to insert more than 1000 documents or more than 16Mb of documents at once. It now automatically splits documents into batches when reaching MongoDB limits (1000pcs or 16Mb).

Fixes #149

### DIFF
--- a/txmongo/connection.py
+++ b/txmongo/connection.py
@@ -16,6 +16,10 @@ from txmongo.database import Database
 from txmongo.protocol import MongoProtocol, Query
 
 
+DEFAULT_MAX_BSON_SIZE = 16 * 1024**2
+DEFAULT_MAX_WRITE_BATCH_SIZE = 1000
+
+
 _PRIMARY_READ_PREFERENCES = {ReadPreference.PRIMARY.mode, ReadPreference.PRIMARY_PREFERRED.mode}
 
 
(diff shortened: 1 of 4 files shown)


txmongo/connection.py
```diff
@@ -16,6 +16,10 @@
 from txmongo.protocol import MongoProtocol, Query
 
 
+DEFAULT_MAX_BSON_SIZE = 16 * 1024**2
```
psi29a: Do we really want to do the calculation here? Why can't we just use: 

DEFAULT_MAX_BSON_SIZE = 16777216  # 16MiB

psi29a:
Besides one comment, looks good to me! :)

IlyaSkriblovsky:
That's just because I like when code is explaining itself without comments :-)  But in this case it's superfluous, agreed.

PR #159, opened 2016-04-27
Change GridFS's default chunkSize to 255kB

touilleMan:
I've also changed `kb` into `KB` (assuming it was a typo)


psi29a:
Thanks! :+1:

PR #161, opened 2016-04-27
Fix GridFS.put yield on close

touilleMan:
Missing `yield` makes `GridFS.put` returns before the file is actually written in GridFS


psi29a:
:+1:

Issue #160, opened 2016-04-27
GridFS.get messing when file is not found


touilleMan:
`GridFS.get` uses `find_one` to retrieve the requested file from the collection.
However nothing is done if the returned value is None, which leads to...

```python
>>> gridout = yield GridFS.get('dummy')
>>> gridout
None
>>> not gridout
False
>>> bool(gridout)
True
>>> type(gridout)
<class 'txmongo._gridfs.grid_file.GridOut'>
```
:trollface:

Of course the returned GridOut object is not usable
```python
>>> gridout.name
*** TypeError: argument of type 'NoneType' is not iterable
```

I think `GridFS.get` should raises a `NoFile` in such case [to be conform with pymongo](https://github.com/mongodb/mongo-python-driver/blob/master/gridfs/__init__.py#L139)

psi29a:
I agree! :)  

If you would like to do this, please do, add a test-case for it and fire off a pull request. Otherwise we'll get around to it eventually.

PR #163, opened 2016-04-27
GridFS.get raises NoFile if file is not found

touilleMan:
As talked in #160 

I've also corrected missing `yield` in `db.fs.files.remove({})` which lead to concurrency errors

### DIFF
--- a/tests/test_objects.py
+++ b/tests/test_objects.py
@@ -179,8 +179,8 @@ class TestGridFsObjects(unittest.TestCase):
         """ Tests gridfs objects """
         conn = yield txmongo.MongoConnection(mongo_host, mongo_port)
         db = conn.test
-        db.fs.files.remove({})  # drop all objects there first
-        db.fs.chunks.remove({})
+        yield db.fs.files.remove({})  # drop all objects there first
+        yield db.fs.chunks.remove({})
         self.assertRaises(TypeError, GridFS, None)
         _ = GridFS(db)  # Default collection
         self.assertRaises(TypeError, GridIn, None)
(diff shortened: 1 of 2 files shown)


touilleMan:
CI failure (`Failure: pymongo.errors.AutoReconnect: TxMongo lost connection to MongoDB.`)
related to #164 : index creation is not over when the database is disconnected

tests/test_objects.py
```diff
@@ -179,8 +179,8 @@ def test_GridFileObjects(self):
         """ Tests gridfs objects """
         conn = yield txmongo.MongoConnection(mongo_host, mongo_port)
         db = conn.test
-        db.fs.files.remove({})  # drop all objects there first
-        db.fs.chunks.remove({})
+        yield db.fs.files.remove({})  # drop all objects there first
+        yield db.fs.chunks.remove({})
```
IlyaSkriblovsky: I suggest using `yield defer.gatherResults(db.fs.files.remove({}), db.fs.chunks.remove({}))` here to not limit concurrency

touilleMan:
Updated PR to use `gatherResults`

IlyaSkriblovsky:
(I've deleted my comment abount `gatherResults` after noticed that it was in tests code and so optimization is not required ;) )

touilleMan:
Well now it's optimized ^^ !

touilleMan:
I've added fix from #161 to make this PR pass the unit tests

tests/test_objects.py
```diff
@@ -258,7 +265,9 @@ def test_GridFsObjects(self):
         conn = yield txmongo.MongoConnection(mongo_host, mongo_port)
         db = conn.test
         gfs = GridFS(db)  # Default collection
-        _ = yield gfs.get("test_3")
+        # Missing file raises error
+        with self.assertRaises(NoFile):
```
IlyaSkriblovsky: Unfortunately, `assertRaises()` can be used as context manager only with Twisted ≥ 15 :(
Please rewrite it as `yield self.assertFailure(gfs.get("test_3"), NoFile)`

touilleMan:
I've changed the PR ;-)

IlyaSkriblovsky:
Looks good, thanks!

psi29a:
:+1:

Issue #164, opened 2016-04-27
GridFS ensure indexes is done without yielding


touilleMan:
https://github.com/twisted/txmongo/blob/master/txmongo/_gridfs/__init__.py#L55

When instanciating a `GridFS` object, an async call to mongo is done to create the chunks collection indexes.
The trouble is this call is never waited so this can lead to errors if the database is disconnected at the wrong time. This is especially true for tests where a lot of conections/disconnections are done in a short time (as a matter of fact I'm surprised this error didn't show up on travis continuous integration...):

```
(py35-twtrunk)➜  txmongo git:(issue-160) ✗ ./.tox/py35-twtrunk/bin/trial  tests.test_objects.TestGridFsObjects.test_GridFsObjects
tests.test_objects
  TestGridFsObjects
    test_GridFsObjects ...                                              [ERROR]

===============================================================================
[ERROR]
Traceback (most recent call last):
  File "/home/emmanuel/projects/txmongo/.tox/py35-twtrunk/lib/python3.5/site-packages/twisted/internet/defer.py", line 1126, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "/home/emmanuel/projects/txmongo/.tox/py35-twtrunk/lib/python3.5/site-packages/twisted/python/failure.py", line 389, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "/home/emmanuel/projects/txmongo/txmongo/collection.py", line 704, in create_index
    yield self._database.system.indexes.insert(index, safe=True, **kwargs)
  File "/home/emmanuel/projects/txmongo/.tox/py35-twtrunk/lib/python3.5/site-packages/twisted/internet/defer.py", line 1126, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "/home/emmanuel/projects/txmongo/.tox/py35-twtrunk/lib/python3.5/site-packages/twisted/python/failure.py", line 389, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "/home/emmanuel/projects/txmongo/txmongo/utils/__init__.py", line 27, in _timeout
    raw_result = yield raw_d
  File "/home/emmanuel/projects/txmongo/.tox/py35-twtrunk/lib/python3.5/site-packages/twisted/internet/defer.py", line 1126, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
[...]

PR #167, opened 2016-04-28
GridFS ensure indexes is done without yielding

touilleMan:
Correction for #164
This PR is based on PR #163 (needed to make the unit tests work), only commit 059355c1a8cb0438c4e46121820fc21895e6fc2c should be considered

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -17,6 +17,9 @@ Features
 
 - ``insert_many()`` is now able to insert more than 1000 documents and more than 16Mb of documents at once.
 - GridFS's default ``chunkSize`` changed to 255kB, to avoid the overhead with usePowerOf2Sizes option.
+- Add ``GridFS.get_indexes_created_defer`` to obtain a defer on the creation of the current
+  GridFS instance's indexes
+- GridFS create indexes for the ``files`` collection in addition to the ``chunks`` one
 
 Release 16.0.1 (2016-03-03)
 ---------------------------
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -17,7 +17,7 @@ Features
 
 - ``insert_many()`` is now able to insert more than 1000 documents and more than 16Mb of documents at once.
 - GridFS's default ``chunkSize`` changed to 255kB, to avoid the overhead with usePowerOf2Sizes option.
-- Add ``GridFS.get_indexes_created_defer`` to obtain a defer on the creation of the current
+- Add ``GridFS.indexes_created`` to obtain a defer on the creation of the current
   GridFS instance's indexes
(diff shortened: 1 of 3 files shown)


IlyaSkriblovsky:
I'd prefer more easy to read name, like `indexes_created()` or `when_indexes_created()`. `get_*` and `*_defer` look like technical details.

Could you please also fix unstable GridFS tests using your new shiny method?

touilleMan:
I've renamed into `indexes_created()`

On the same subject, I've seen that pymongo [has stopped writting indexes when instanciating `GridFS`](https://github.com/mongodb/mongo-python-driver/blob/master/gridfs/__init__.py#L51)
It is now done on first `write`. While we are into this PR, do you think it worth copying pymongo behavior on this ?

IlyaSkriblovsky:
Actually, I'm not familiar with GridFS at all. @psi29a might have an opinion on that.

psi29a:
I'm also not familiar with GridFS, but we do have a policy of trying to be as 1:1 with PyMongo where possible.

touilleMan:
@psi29a I'll modify this PR to reflect pymongo behavior then

touilleMan:
Digging a bit on this, I've found a trouble:
PyMongo way is now do the indexes check/creation inside the first call to `GridIn.__flush_data`. 

However it doesn't seems a really good idea for performances given the indexes will be enforced for every `GridIn` object flushed (i.e. for every file uploaded) instead of doing it once per `GridFS` instanciation. I guess the strategy is to assume checking for indexes is cheap and this save trouble with concurrent indexes drop.

Considering TxMongo, this is not optimal either given there would be an indexe check/creation defer for each `GridIn` instance (some of them being even hidden from the user, like when we use `GridFS.put` !).

A solution would be to provide to provide a callback to the `GridIn.__init__` to register and collect into the `GridFS` all the defers. However this feels pretty cumbersome so I'm not sure we should do it or stick with the current behavior...

@psi29a & @IlyaSkriblovsky what do you think ?

IlyaSkriblovsky:
I'm personally don't think this is a big deal. I think ensuring indexes in `GridFS.__init__` is good enough.

P.S. Just can't decide should I write "indices" to be correct or "indexes" to be easily read 😕

touilleMan:
Given the[ pymongo documentation](https://api.mongodb.org/python/current/api/pymongo/collection.html#pymongo.collection.Collection.create_indexes) talks about "indexes" I think we should keep with it

touilleMan:
Any progress on this ?

IlyaSkriblovsky:
Sorry, we probably missed it. Could you please rebase this to master to resolve conflicts?

touilleMan:
rebased

psi29a:
Looks great! Thanks! :+1:

PR #168, opened 2016-04-28
fields parameter removed for count()

IlyaSkriblovsky:
`fields` projection parameter seems to be completely pointless for `count()`

I don't think removing it could break any existing code.

Also added distinct test case for `count()`


psi29a:
I agree with this. Thanks!

PR #171, opened 2016-06-21
default watchdog timeouts increased

IlyaSkriblovsky:
It turns out that low default values prevents us from issuing long-running queries


glyph:
Looks like a good change to me.  It should be possible to run a query longer than a few seconds.  I hope you don't mind if I approve and merge!

psi29a:
It's fine. Though we still need a better solution right @IlyaSkriblovsky ?

IlyaSkriblovsky:
@psi29a, yep. Sorry for the delay, I really had no chance yet to work on it.

PR #172, opened 2016-06-22
Collection.bulk_write()

IlyaSkriblovsky:
Implementation of PyMongo's `Collection.bulk_write()` method.

Some internals of PyMongo used to avoid copy-pasting.

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -1,6 +1,16 @@
 Changelog
 =========
 
+Release 16.2.0 (UNRELEASED)
+---------------------------
+
+Features
+^^^^^^^^
+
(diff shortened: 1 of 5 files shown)


psi29a:
Well that was fast. ;)

psi29a:
LGTM :+1: 

Can you update the changelog/news file please?

This is worthy of a 16.2 release ;)

IlyaSkriblovsky:
Updated NEWS.rst. Sorry, I'm always forgetting about that.

I think we should handle watchdog issue before 16.2.0... At least increase default timeouts as suggested in #171.

psi29a:
Done, thanks! :)

PR #173, opened 2016-06-23
Fixed handling of connection errors in bulk_write

IlyaSkriblovsky:
`bulk_write()` was not propagating exceptions (`AutoReconnect`, for example) when used with `ordered=False`


psi29a:
Glad you caught it!

PR #177, opened 2016-07-20
support for codec_options

IlyaSkriblovsky:
We have used `pymongo.bulk._Bulk` internal class to implement bulk ops. But since PyMongo 3.3.0 that was released a week ago, this class assumes that `Collection` has proper support of codec options.

This patch adds `codec_options` to `ConnectionPool`, `Database` and `Collection` classes and `codec_options` kwarg to `Collection.with_options()`. This fixes our test suite that is failing since PyMongo 3.3 release.

I've also added notice in NEWS.rst that `Collection.with_options` should be preferred now instead of `as_class` kwags to `find`. PyMongo had dropped `as_class` kwags in 3.0.

PR #178, opened 2016-07-21
fixing tests against twisted-trunk

IlyaSkriblovsky:
Twisted team changed something in either Trial or `twisted.python.reflect` and now our `trial tests` says `can't find module "tests"`. I'm not sure if it is a bug in Twisted and whether it will be fixed, but lets fix this on our side for now.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -17,6 +17,8 @@ deps =
     tw153: Twisted==15.3
     tw150: Twisted==15.0
     tw140: Twisted==14.0
+setenv =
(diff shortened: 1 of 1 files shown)


IlyaSkriblovsky:
Merging despite test failures because now failures is only due to #177, which I'm going to rebase onto this PR to makes tests to pass finally

psi29a:
Was out the past couple of days, glad you found the fix. I'll likely have to do this with Ldaptor as well.

PR #179, opened 2016-08-01
Added support for text indexing.

jdcumpson:
Allow text indexes. 

Slight format change in query filters to allow the class to be more
extensible in the future with class variable `ALLOWED_DIRECTIONS`.


psi29a:
Great! :)
Can you add a test case or two as well please?

jdcumpson:
Certainly! Thanks for the feedback.

IlyaSkriblovsky:
Transformed into #188 and merged

PR #180, opened 2016-08-14
Various tests fixes

IlyaSkriblovsky:
Small tests improvements:
• `test_replicaset.py` was using 3 GB of temporary disk space (1GB oplog per 3 `mongod` instances)
• `test_auth.py` might leak directories in `/tmp` when ran against MongoDB v3
• Not testing with py26
• Test all supported configurations: py27 + Twisted ≥14, py3x + Twisted ≥16, pypy + Twisted ≥14

But the number of test configurations doubled, which will make Travis tests to run longer

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,15 +16,31 @@ python: 2.7
 #    - pypy
 
 env:
+    - TOX_ENV=py27-tw140
+    - TOX_ENV=py27-tw150
+    - TOX_ENV=py27-tw155
+    - TOX_ENV=py27-tw160
+    - TOX_ENV=py27-tw163
     - TOX_ENV=py27-twlatest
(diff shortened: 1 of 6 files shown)


IlyaSkriblovsky:
Rebased onto master branch and also fixed testing against MongoDB 3.2.

I think it is ready for review

psi29a:
Looks good to me aside from some nested try/except blocks that I'm iffy about. ;)

IlyaSkriblovsky:
They were already there before this patch :)

They are nested because I need to run `except` block *after* `finally`. I don't see a good way to refactor this except move inner try-blocks into separate methods. Should I do this?

psi29a:
No need, this is test code anyway. ;)

PR #181, opened 2016-08-15
Pinging server(s) and drop connections if server is gone

IlyaSkriblovsky:
*This PR is not for merging now, but mostly for discussion of the issue and the ways of solution*

Checking regularly that server nodes that we are connected to are alive and dropping connections if they are not.

This is one of possible approaches to solve the issue: if the network link is disappears between us and MongoDB server, how can we deal with it?

Currently TxMongo regularly sends `ismaster` command to all connections in the pool and drops connection if no answer received within a timeout. But this has a downside: if connection is waiting for a very long-running query, it won't receive `ismaster` response and will be dropped by this watchdog procedure.

What do you think about suggested technique? Does it cover all the scenarios? Can the code be simplified?

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -20,6 +20,9 @@ API Changes
 ^^^^^^^^^^^
 
 - ``Database.command()`` now takes ``codec_options`` argument.
+- ``watchdog_interval`` and ``watchdog_timeout`` arguments of ``ConnectionPool`` renamed
+  to ``ping_interval`` and ``ping_timeout`` correspondingly along with internal change of
+  connection aliveness checking mechanism.
 
 Bugfixes
 ^^^^^^^^
--- a/tests/test_cancel.py
+++ b/tests/test_cancel.py
@@ -68,7 +68,7 @@ class TestCancelParts(unittest.TestCase):
     @defer.inlineCallbacks
     def test_connection_notifyReady(self):
         uri = parse_uri("mongodb://localhost:27017/")
-        conn = _Connection(None, uri, 0, 10, 10, 10, 10)
+        conn = _Connection(None, uri, 0, 10, 10)
         d1 = conn.notifyReady()
         d2 = conn.notifyReady()
         d1.cancel()
--- a/tests/test_replicaset.py
+++ b/tests/test_replicaset.py
@@ -279,8 +279,7 @@ class TestReplicaSet(unittest.TestCase):
 
     @defer.inlineCallbacks
     def test_StaleConnection(self):
-        conn = MongoConnection("localhost", self.ports[0],
-                               watchdog_interval=10, watchdog_timeout=5)
+        conn = MongoConnection("localhost", self.ports[0], ping_interval = 5, ping_timeout = 5)
         try:
             yield conn.db.coll.count()
             self.__mongod[0].kill(signal.SIGSTOP)
--- a/tests/test_replicaset.py
+++ b/tests/test_replicaset.py
@@ -282,6 +282,9 @@ class TestReplicaSet(unittest.TestCase):
         conn = MongoConnection("localhost", self.ports[0], ping_interval = 5, ping_timeout = 5)
         try:
             yield conn.db.coll.count()
+            # check that 5s pingers won't break connection if it is healthy
+            yield self.__sleep(6)
+            yield conn.db.coll.count()
             self.__mongod[0].kill(signal.SIGSTOP)
             yield self.__sleep(0.2)
             while True:
--- a/txmongo/connection.py
+++ b/txmongo/connection.py
@@ -10,11 +10,10 @@ from pymongo.uri_parser import parse_uri
 from pymongo.read_preferences import ReadPreference
 from pymongo.write_concern import WriteConcern
 from twisted.internet import defer, reactor, task
-from twisted.internet.protocol import ReconnectingClientFactory
+from twisted.internet.protocol import ReconnectingClientFactory, ClientFactory
 from twisted.python import log
 from twisted.python.compat import StringType
 from txmongo.database import Database
-from txmongo.errors import TimeExceeded
 from txmongo.protocol import MongoProtocol, Query
 from txmongo.utils import timeout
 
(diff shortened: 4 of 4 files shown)


psi29a:
Those are a lot of failures...

IlyaSkriblovsky:
Sorry, messed up with stopping that `LoopingCall`. What do you think about general idea? Does it sound reasonable?

psi29a:
OK, so this still relies on ismaster which can be blocked on long queries right?

IlyaSkriblovsky:
Yes, this implementation sends `ismaster`, but it sends it in another, newly created connection, so it won't be blocked (long queries only block other queries in same connection (except for writes)).

Btw, currently it creates new connection every timer tick. This connection can probably be cached between ticks, but then if ping is failed, we need to re-check with new connection.

psi29a:
Gotcha, so this is the N+1 scenario we talked about. The +1 connection will never be exposed to the pool for use by txmongo clients.

Every timer tick is what? Default ping interval (30s)?

IlyaSkriblovsky:
> Every timer tick is what? Default ping interval (10s)?
Yes, `ping_interval=10`. But timing defaults are the subject of discussion, of course

Every `ping_interval` we open new connection (not in pool) to *each* distinct address of server nodes we are connected too (because with `slave_ok=True` connections in the pool might be to different nodes). In these new connections we send `ismaster` and if we received any answer within `ping_timeout` seconds, we consider this server address alive. If the procedure fails for any address, we are forcibly closing all connections in the pool that are connected to *this* address.

psi29a:
Ah, so it is our total number of connections is N*2 of which N is our pool size. Is there an impact on performance or system load by constantly opening and closing a connection to MongoDB?  Is there a reason for doing this instead of having a long-living connection just to run ismaster per connection in the pool?

IlyaSkriblovsky:
> Ah, so it is our total number of connections is N*2 of which N is our pool size

Yes. More presicely: N+1 for `slaveOk=False` (the default) and up to N+M where M is replicaset size for `slaveOk=True`.

>  Is there a reason for doing this instead of having a long-living connection

Agreed, long-living connections seem more reasonable. But if `slaveOk=True` we will need to discover *all* primary and secondary nodes and have a persistent connection to each one.

> Wouldn't it be interesting to use this ismaster as our heart-beat to handle if ismaster is false cases too? 

Actually, I don't like our existing heartbeat technique... If it doesn't give us a precise result, why do we need it at all?

psi29a:
I agree!  The management connections won't be blocked by long running queries.

IlyaSkriblovsky:
Ok, I will try to rewrite this using persistent connections. But it will probably require additional topology discover process, because now only Connections know full node list, not ConnectionPool.

How do you think: if `ismaster` via persistent connection will fail to respond in a given timeout, is it enough condition to drop all pool connections to this address?

psi29a:
If the persistent "management" connection fails to give us an ismaster within a short period of time, then there is something very wrong. You've already proven that ismaster will be blocked by a long running query, but only if running on the same connection. I believe that this is one of the most important jobs in TxMongo when dealing with a replicaset.

IlyaSkriblovsky:
@psi29a I mean, is it possible that management connection is somewhy stalled while normal pool connections are still alive? Do we need to recheck ismaster twice before dropping pool connections? I'm not sure

psi29a:
ah, a double-check: "we got reports of an outage, let's check the real connection". 

Then the double-check should happen but with a even smaller timeout than the first.

IlyaSkriblovsky:
@psi29a I don't think we should re-check using real connection from the pool. Because that pool connection might be blocked by long-running query. I meant re-checking via new maintaince connection. If check is failed, drop that maintaince connection, establish new one instead and re-check. Does it sound reasonable? Is there any real benefit from re-checking? I would be happy to keep this logic as simple as possible...

psi29a:
It does sound reasonable, sorry if this was blocked by me.  I thinking keeping this simple is key to minimizing weird issues in the future.

IlyaSkriblovsky:
It was blocked by lack of spare time of mine :)

IlyaSkriblovsky:
I've rewritten availability checking code using persistent connections. Please have a look on the code when you have time.

The mechanics is generally described in large comment in the code. The idea is that we maintain one pinger connection per address, periodically send `ismaster` via each connection and drop all connections to specific address if `ismaster` fails. If we fail to establish maintaince connection, all pool connections are dropped too. If maintaince connection got disconnected for some reason, we doesn't drop pool connections. In that case maintaince connection will be reestablished by periodic `__discovery_nodes_to_ping`.

psi29a:
This looks OK to me. We ready to commit to this? :)

IlyaSkriblovsky:
No, I think we should also remove old watchdog code that sends `ismaster` inside every pool connection and drops it if answer is not received withing X seconds. As we discussed earlier, this approach can break  long-running queries. New method, implemented in this PR is free from this side effect.

So, if you are ok with this implementation, I will remove the old method if you don't mind.

psi29a:
Absolutely, the old watchdog code should be removed.  The work you've done is fantastic and I think this is the best way forward.

IlyaSkriblovsky:
I've removed old watchdog code and fixed "unclean reactor" test failures. It is ready for final review and merging.

psi29a:
No real comments aside from the fact that this is very clean, a proper job!

psi29a:
Perhaps an update to the changelog?

IlyaSkriblovsky:
@psi29a Aah, sorry again, Mr. Changelog!

IlyaSkriblovsky:
1. What do you think about default values of `ping_interval` and `ping_timeout`? They are 10s both now. Does it sounds reasonable?
2. Previous buggy watchdog code used `watchdog_interval`/`watchdog_timeout` arguments to `ConnectionPool` constructor. New code uses `ping_interval`/`ping_timeout`. May be better to leave old names and not break compatibility? The meaning of these arguments are pretty similar anyways.

psi29a:
I'm comfortable with the ping_interval and ping_timeout  in terms of default values and their names. I think most people know what ping is and the watchdog_* names are introduced only recently, if we get it out for a new release before the new year, we should be OK I think. 

I'm going to assume (uh oh!) that most people use the default values anyway (as in don't know the parameters exist to begin with).

You ready for merge? :)

IlyaSkriblovsky:
I just started to write changelog entry when question about parameter names arose. If we leave them as `watchdog_*`, no visible for users API changes need to be documented. But OK, lets rename them to `ping_*`, I like it more too :)

IlyaSkriblovsky:
changelog entry added!

PR #183, opened 2016-09-29
Multi-key sort specs and hints fixed

IlyaSkriblovsky:
TxMongo has the serious bug: if sorting specification or hint contains several fields, it gets transformed to a `dict` before issuing the query, so the real fields order in BSON is undefined and so ordering is undefined too.

This patch fixes it by using `SON` instead of `dict`. Unfortunately, I can't imagine how to make an unintrusive test case for this except making `Collection.__apply_find_filter` non-private and test by calling it.

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -13,6 +13,11 @@ Features
 - ``codec_options`` properties for ``ConnectionPool``, ``Database`` and ``Collection``.
   ``Collection.with_options(codec_options=CodecOptions(document_class=...))`` is now preferred
   over ``Collection.find(..., as_class=...)``.
+  
+Bugfixes
+^^^^^^^^
+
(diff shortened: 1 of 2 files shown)


psi29a:
Can you add this (and other changes) into the changelog as well?

IlyaSkriblovsky:
Done. Btw, this bug seems to be serious enough to bump the version with bugfix.

I'm especially worried because this bug was introduced by me :facepalm:

IlyaSkriblovsky:
Is it OK to merge this?

psi29a:
Sounds good to me, thanks!

IlyaSkriblovsky:
@psi29a Don't you think it's worth to release this bugfix? It seems to me critical enough

psi29a:
On it :)

psi29a:
Uploaded to pypi: 16.2.0

PR #184, opened 2016-09-29
Tests fixed when running with WiredTiger

IlyaSkriblovsky:
`Collection.index_information` and `Collection.options` are depending on `db.system.indexes` and `db.system.namespaces` which are not exist when running on WiredTiger.

This patch fixes `index_information()` and `options()` by using new-in-3.0 `listIndexes` and `listCollection` commands.

Both new commands return results in the new-style cursor specification document which we are not able to properly unwind (yet). So, `assert`-ing that all results are coming in one batch. I think this is ok because it is hard to imagine the amount of indexes or collection whose info wouldn't fit in 16Mb.


psi29a:
LGTM, thanks!

PR #189, opened 2016-10-04
Implementation of 'get_version'

oddjobz:
This is based on the PyMongo call available in the v3.4 API and is pretty critical for anything that makes use of Mongo's versioning feature. I've added an additional index to ensure you can scan both-ways in the version list, and I've added a test-case to exercise the new code. The PyMongo implementation also allows random (potentially/probably non-indexed) attributes to be included in the search, I have deliberately omitted this feature as it's "likely" to introduce unpredictable performance issues if it's ever used. (although it anyone needs it, adding is relatively trivial) I don't necessarily "like" the indexing scheme, but it mirrors the PyMongo implementation for compatibility.


docs/source/NEWS.rst
```diff
@@ -1,6 +1,18 @@
 Changelog
 =========
 
+Unreleased
```
IlyaSkriblovsky: We are changed NEWS.rst today by adding unreleased 16.3.0 version. Could you please merge current `master` branch of txmongo into your branch and resolve the merge conflict?

tests/test_get_version.py
```diff
@@ -0,0 +1,81 @@
...
+            yield self.gfs.put("Hello" + str(n), filename="world")
+
+    @defer.inlineCallbacks
+    def tearDown(self):
+        docs = yield self.gfs.list()
+        self.assertEqual(docs, ['world'])
+        index = 9
+        while True:
+            try:
+            	doc = yield self.gfs.get_last_version('world')
```
IlyaSkriblovsky: Please, don't mix tabs and spaces in indents :)

txmongo/_gridfs/__init__.py
```diff
@@ -138,6 +141,45 @@ def get(self, file_id):
         defer.returnValue(GridOut(self.__collection, doc))
 
     @defer.inlineCallbacks
+    def get_version(self, filename=None, version=-1, **kwargs):
```
IlyaSkriblovsky: If you have decided to omit custom filtering feature, lets drop `**kwargs` here too in order to not make an illusion that it will work as in PyMongo

txmongo/_gridfs/__init__.py
```diff
@@ -138,6 +141,45 @@ def get(self, file_id):
...
+            skip -= 1
+            myorder = DESCENDING("uploadDate")
+        else:
+            myorder = ASCENDING("uploadDate")
+
+        cursor = yield self.__files.find(query, filter=filter.sort(myorder), limit=1, skip=skip)
+
+        try:
+            defer.returnValue(GridOut(self.__collection, cursor[0]))
+        except StopIteration:
```
IlyaSkriblovsky: `yield find()` will return plain list, not cursor object like in PyMongo. So `IndexError` should be cought here, not `StopIteration`. Or better just checking:
```python
if cursor:
    defer.returnValue(...)
raise NoFile(...)
```

tests/test_get_version.py
```diff
@@ -0,0 +1,81 @@
...
+    @defer.inlineCallbacks
+    def setUp(self):
+        self.conn = connection.ConnectionPool("mongodb://127.0.0.1/dbname")
+        self.db = self.conn['dbname']
+        self.gfs = gridfs.GridFS(self.db)
+        for n in xrange(0,10):
+            yield self.gfs.put("Hello" + str(n), filename="world")
+
+    @defer.inlineCallbacks
+    def tearDown(self):
```
IlyaSkriblovsky: This `tearDown` looks like a full-weight test :)  May be extract this loop into distinct test method and leave only `disconnect` (and may be `self.db.drop()`)?

IlyaSkriblovsky:
Sorry, I've misguided you about the way to drop database. Currently, there is no any `drop_database` method in TxMongo. You may drop your database with `self.db.command("dropDatabase")`.

oddjobz:
Ok, I was a little confused by the drop issue, looks ok now. Indeed the new version of the test code exercises the StopIteration fix .. :) .. how's it look?

tests/test_get_version.py
```diff
@@ -0,0 +1,89 @@
...
+        yield self.db.command('dropDatabase')
+        yield self.conn.disconnect()
+
+    @defer.inlineCallbacks
+    def test_GFSVersion_strip(self):
```
IlyaSkriblovsky: Seems like this test is testing `get_last_version()` which was there for a long time, not `get_version()` which you have added. May be `get_last_version('world')` should be replaced with `get_version('world, -1)`?

tests/test_get_version.py
```diff
@@ -0,0 +1,89 @@
...
+        while True:
+            try:
+                doc = yield self.gfs.get_last_version('world')
+                text = yield doc.read()
+                self.assertEqual(text, 'Hello' + str(index))
+                index -= 1
+                yield self.gfs.delete(doc._id)
+            except NoFile:
+                break
+        try:
```
IlyaSkriblovsky: Do we need this `try` block? If we got out of `while True` loop, it certainly mean that `yield self.gfs.get_last_version('world')` raises the `NoFile`. Why re-check?

I guess you have added this block to check `StopIteration` fix, but it doesn't actually test it because it calls `get_last_version`, not `get_version` :)

tests/test_get_version.py
```diff
@@ -0,0 +1,89 @@
...
+            doc = yield self.gfs.get_last_version('world')
+            self.assertEqual(doc, None)
+        except NoFile:
+            return
+
+    @defer.inlineCallbacks
+    def test_GFSVersion_last(self):
```
IlyaSkriblovsky: This test seems to be calling `get_last_version` instead of `get_version(..., -1)` too

IlyaSkriblovsky:
I've updated my review. Just a couple of minor comments about testing code :)

tests/test_get_version.py
oddjobz: Yup,fixed.

tests/test_get_version.py
oddjobz: Sure, that was from me trying to figure out why my "drop" was failing (silently), but you're right, it's not needed. (removed)

tests/test_get_version.py
oddjobz: I do have 1 last "last" call left, call it "belt and braces", not strictly needed but nice to make sure get_last and get_version(-1) give the same result .. :)

IlyaSkriblovsky: Ah, ok, I just didn't noticed that this test is actually duplicated: one calls `get_version` and another calls `get_last_version`

oddjobz:
Ok, comment should read ".. remove get_last_version" .. should look better now ...

txmongo/_gridfs/__init__.py
```diff
@@ -62,6 +62,9 @@ def __init__(self, database, collection="fs"):
         self.__indexes_created_defer = defer.DeferredList([
             self.__files.create_index(
                 filter.sort(ASCENDING("filesname") + ASCENDING("uploadDate"))),
+            self.__files.create_index(
+                filter.sort(ASCENDING("filesname") + DESCENDING("uploadDate"))),
```
IlyaSkriblovsky: After careful reading I've noticed that second index is redundant. `{filename: 1, uploadDate: 1}` index will cover both `find({filename: ...}).sort({uploadDate: 1})` and `find({filename: ...}).sort({uploadDate: -1})`.

IlyaSkriblovsky:
The last comment, I beleive :)

txmongo/_gridfs/__init__.py
oddjobz: Ok, in my defence I was just copying other people's code with regards to index creation, and if one of those indexes is obsolete, it would imply the current method for specifying index creation is also rather redundant. I've done a little reading too now ( :) ) so here's another shot with one index, with a slightly more direct specification, and also I modified the test to it runs under both Python2 "and" Python3.

oddjobz: /s/test to/test so/

IlyaSkriblovsky:
Nice catch about Py3 compatibility!

But please, revert back index creation code:
```python
self.__files.create_index(
    filter.sort(ASCENDING("filesname") + ASCENDING("uploadDate")))
```

`create_index({'filesname':1, 'uploadDate':1}` is not correct because python's dicts do not guarantee key order, so you may get `{uploadData:1, filesname:1}` index instead. `filter.sort(ASCENDING(...) + ASCENDING(...))` is the correct way to specify indexes, though a little bit verbose.

Still you are correctly removed `ASCENDING(...) + DESCENDING(...)` index, thanks.

txmongo/_gridfs/__init__.py
```diff
@@ -60,7 +60,7 @@ def __init__(self, database, collection="fs"):
         self.__files = self.__collection.files
         self.__chunks = self.__collection.chunks
         self.__indexes_created_defer = defer.DeferredList([
-            self.__files.create_index(
+        self.__files.create_index(
```
IlyaSkriblovsky: You are probably already tired of my grumling, but could you please fix indenting of this line? ;)

oddjobz:
Good point, reverted .. that's still a horribly clunky way to define an index .. would be nice to support something a little more intuitive .. maybe [{'field1':1},{'field2':2}] .. the way the Mongo guys implemented index spec's is a little unfortunate with regards to Python dict's ... :(

oddjobz:
Ok, turns out this would be a valid index spec;
```python
create_index([('filesname':1),('uploadDate':1)], name='version_index')
```
I would submit this is 'clearer' in terms of what it's doing ... (?)

IlyaSkriblovsky:
You can specify indexes like this: `[("field1", 1), ("field2", 1)]` and it will work. `filter.sort` and `filter.ASCENDING` are internally converted into exactly this. That verbose notation is just a traditional way of specifying indexes.

I suggest you to leave it as it was, with all these `filter.sort` and `ASCENDING`, because it turns out that this PR doesn't really need to change anything about indexes. It may probably be redone in clearer way by another pull request.

Sorry, but it seems that your last commit actually broke indenting of another line instead of fixing :)  Both indexes are created inside `DeferredList` constructor, please make them indented one level more than line with `DeferredList` call. Thanks!

oddjobz:
Ok, how's that?

oddjobz:
:)

IlyaSkriblovsky:
Let's wait for results of automated testing by Travis

IlyaSkriblovsky:
Thanks for your contribution!

psi29a:
Yes, thank you very much and congratulations! :D

PR #190, opened 2016-10-04
GridFS index fixed. codec_options argument for Database.command.

IlyaSkriblovsky:
It is funny and sad at the same time, but GridFS were creating index with a typo in field name and nobody noticed that :)

Also, `ensure_index` removed from `get_last_version` because ASC+ASC index, created in `__init__` is sufficient.

For the means of testing indexes creation, I've added `codec_options` argument to `Database.command()`. It can be used to make `command()` to return response as `SON` instead of plain `dict`.

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -16,6 +16,16 @@ Features
   as it may create *very* variable conditions in terms of loading and timing. An additional index is		
   also added to facilitate bi-directional movement between versions. Additional Unit test for GFS also added.
 
+API Changes
+^^^^^^^^^^^
+
+- ``Database.command()`` now takes ``codec_options`` argument.
+
+Bugfixes
+^^^^^^^^
+
+- ``GridFS.get_last_version()`` was creating redundant index
+
 Release 16.2.0 (2016-10-02)
 ---------------------------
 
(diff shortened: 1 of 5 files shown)


psi29a:
We now have conflicts... to be expected now with less inline callbacks.

IlyaSkriblovsky:
Rebased and solved conflicts.

I've noticed also that adding `codec_options` to `Database.command` was unnecessary since it passes all its `**kwargs` to `find_one()`, so I just can use `as_class=SON`. But `as_class` is deprecated in PyMongo for a long time and `codec_options` just reflects PyMongo's API, so I suggest to leave it.

P.S.
(Btw, this relates to the discussion of hidden `**kwargs` in another PR ;) )

P.P.S.
More on `**kwargs`. Notice how I had to do `command()` call here: https://github.com/twisted/txmongo/blob/master/txmongo/collection.py#L193
I can't just write `command("listCollections", filter={"name": self.name})` because `command()` uses kwargs twice: it adds them to command payload and *also* passes them to `find_one()`. So `"filter"` kwarg not only adds up to DB command, but also messes it up on lower level. I hope we eventually will clean and tidy all these kwargs magic :)

IlyaSkriblovsky:
@psi29a how do you feel you about this PR now?

psi29a:
LGTM, merging.

psi29a:
Agreed about the kwarg magic, it's dark band-aid stuff that we should move away from. I formally apologize for adding to it. ;)

PR #191, opened 2016-10-08
Removing most of inlineCallbacks for the sake of cancellation support

IlyaSkriblovsky:
`inlineCallbacks` is totally cool with one drawback: it doesn't properly supports `.cancel()`. If you cancel `inlineCallbacks`-based function it will silently continue to work and just ignore the result. Currently there is no any way for external code to stop `inlineCallbacks` in the middle of its execution.

On the other hand, plain old Deferreds are rather good at cancellation: when all the call stack is written in plain `addCallback`-style, `.cancel()` is able to stop execution at any stage without any explicit support from the code.

So, I suggest to get rid of almost all `inlineCallbacks` in txmongo stack to make it possible to cancel all DB operations.

It should be noted though that `cancel()` will have limited usefulness because in normal operation txmongo sends requests to DB immediately and such queries can't be really cancelled. `.cancel()` can only prevent execution of queries that was held back by waiting for active connection in the pool or complex multi-stage operations like query with getmore's or bulk writes.

Other minor advantage of removing `inlineCallbacks` is that it takes considerably more RAM, up to ×3 times according to my benchmarks. In app with tons of concurrent queries it may result in performance increase under high loads (high memory usage calls GC more often).

This PR changes a lots of code and also replaces some nice `yield`-based loops with tricky recursive callbacks. **So it needs careful code review!**

If this PR will be merged, we can eventually get rid of `@timeout` and `check_deadline()` in favor of upcoming `Deferred.addTimeout()` (https://github.com/twisted/twisted/commit/f4b3869acfb354c8ad62b44b79da3c8cac80e9af)

Remaining issues:
- `find()` should kill cursor if cancelled between loading result batches

Questions to discuss:
- I've added simple cancellers for Deferreds, created by `txmongo.protocol` and `txmongo.connection` that simply unregister deferred from the waiting lists. Such cancellers are actually useless, because they are effectively identical to Deferreds without cancellers (because Deferred without canceller have special semantics: they are just swallowing `.callback()` call after cancellation and doesn't call the callback chain, so effect would be the same). Should I keep them or its better to remove them?
- Functions that were `inlineCallback`-based, previously returned failed deferreds if arguments didn't passed checks. They are now raising argument checking errors directly, without wrapping with `Deferred`. Hence some `assertFailure` was replaced by `assertRaises` in testing code. This is actually the backwards-incompatible change. Should I wrap these `TypeErrors` with `defer.fail()` to make code backwards-compatible?


trenton42:
I think this is a really good change. I haven't reviewed the code yet, but It seemed to me there were a lot of `inlineCallbacks` added for no reason at all. Quite a daunting task to day on, so high five to @IlyaSkriblovsky for taking that on!

IlyaSkriblovsky:
@trenton42 Thanks! Code review would be very appreciated. I'm afraid I might be messing something with complex cases like bulk_write and related functions.

psi29a:
Looks like something went bad in the last commit...

I'll give this a go over today at lunch.  Same with other PR. 

I would still like for there to be timeouts and deadlines, I'm hoping this will just help simplify the code base?

IlyaSkriblovsky:
@psi29a, aah, sorry, there was a stupid mistake in last commit. Fixed.

Actualy, I don't much like current timeout implementation for two reasons:

1. It adds hidden implicit `**kwargs` to every query method. Such hidden arguments are not visible in the method signature, they are not handled properly by IDEs, they are error-prone for typos, etc. I think we should eventually eliminate almost all `**kwargs` and replace them by explicit keyword arguments. `Explicit is better than implicit` :)  By the way, PyMongo already did it: before we were specifying write-concern, BSON options like `as_class` and other options via implicit kwargs which were passed back and forth thru PyMongo's internal functions. But in 3.x they moved all these general query options to explicit clean `with_options` and all new API methods introduced in 3.x are without kwargs.

2. Timing out Deferreds is a common task. Why do it inside txmongo when it can be done outside? Upcoming release of Twisted will introduce `Deferred.addTimeout()` which does roughly the same. So `coll.find(..., timeout=10)` might be converted to `coll.find(...).addTimeout(10, reactor)` and the latter seems cleaner and more universal. There is no `addDeadline()` though but it can be easily done with standalone function rather than mixing it inside every query method. `@timeout` was implemented that way because txmongo didn't supported cancellation and support was needed from `check_deadline()` calls inside query methods implementation. If this PR will be merged, same effect can be achieved by simply cancelling returned deferreds, without any special support from txmongo's internals. So, the task of timing out can be separated from txmongo's internals.

Anyway, even if you want to remain the current API, `@timeout` implementation could be shortened like this:
```python
def timeout(func):
    def _timeout(*args, **kwargs):
        d = func(*args, **kwargs)
        timeout = calc_timeout_from_kwargs(kwargs)
        if timeout:
            d.addTimeout(timeout)
        return d
     return _timeout
```

IlyaSkriblovsky:
Tests against Twisted trunk are failing because today they have changed attributes of `twisted._version` (https://github.com/twisted/twisted/commit/5c1777777f85aaf799c2269f9f5dddcf0d59803e). I will create another PR to address this.

psi29a:
The reasons why it was done inside of txmongo was because canceling deferreds with @inline.callbacks was an impossibility. ;) Chicken/Egg problem...

How much older do you mean, in terms of Twisted?  Are we OK with Twisted >= 15.0

psi29a:
In theory, now that we have cancelable deferreds, a lot of the timeout/deadline code can go away... along with that nasty hack.

IlyaSkriblovsky:
This PR is meant to remove the Chicken :)

`addTimeout` is only in Twisted trunk, it is not released yet. So, I'm speaking mostly about future. Sorry, I wasn't clear about that :)

IlyaSkriblovsky:
Currently, after accepting this PR we might only simplify the implementation of `@timeout` a bit by removing `check_deadline()` and hidden `"_deadline"` kwarg.

IlyaSkriblovsky:
Rebased

tests/test_collection.py
```diff
@@ -67,17 +67,21 @@ def make_col(base, name):
...
-        self.assertFailure(self.db.test.insert(1), TypeError)
-        self.assertFailure(self.db.test.update(1, 1), TypeError)
-        self.assertFailure(self.db.test.update({}, 1), TypeError)
-        self.assertFailure(self.db.test.update({}, {}, 'a'), TypeError)
+        self.assertRaises(TypeError, self.coll.save, "test")
+        self.assertRaises(ValueError, self.coll.filemd5, "test")
+        #self.assertFailure(self.db.test.find(spec="test"), TypeError)
+        #self.assertFailure(self.db.test.find(fields="test"), TypeError)
+        #self.assertFailure(self.db.test.find(skip="test"), TypeError)
+        #self.assertFailure(self.db.test.find(limit="test"), TypeError)
```
psi29a: Why are these commented?

txmongo/collection.py
```diff
@@ -181,39 +181,44 @@ def _normalize_fields_projection(fields):
...
+            assert response["cursor"]["id"] == 0
+            first_batch = response["cursor"]["firstBatch"]
+            if first_batch:
+                return first_batch[0]
+            else:
+                return None
+
+        return self._database.command(SON([("listCollections", 1),
+                                     ("filter", {"name": self.name})]))\
+                   .addCallback(on_ok)
```
psi29a: For readability, can you do a carriage return before SON to get it to the next line? Will this give us enough room to attach .addCallback(on_ok) without having to append an \  ?

tests/test_collection.py
IlyaSkriblovsky: They are replaced by `assertRaises` below. Commented ones should be removed, just forgot it.

txmongo/collection.py
IlyaSkriblovsky: done

tests/test_connection.py
```diff
@@ -64,11 +64,9 @@ def test_Timeout_and_Deadline(self):
         yield self.named_conn.db.coll.insert({'x': 42}, safe=True, timeout=10)
         yield self.named_conn.db.coll.insert({'x': 42}, safe=True, deadline=time()+10)
 
-        d_insert = self.named_conn.db.coll.insert({'x': 42}, safe=True, deadline=time()-10)
-        yield self.assertFailure(d_insert, TimeExceeded)
+        self.assertRaises(TimeExceeded, self.named_conn.db.coll.insert, {'x': 42}, safe=True, deadline=time()-10)
 
-        d_insert = self.named_conn.db.coll.insert({'x': 42}, safe=True, timeout=-10)
-        yield self.assertFailure(d_insert, TimeExceeded)
+        self.assertRaises(TimeExceeded, self.named_conn.db.coll.insert, {'x': 42}, safe=True, timeout=-10)
```
IlyaSkriblovsky: I'm not sure... Previously all exceptions from `@inlineCallbacks`-based methods were returned as failed Deferred. And calling code might be sure that it will always return Deferred. This PR changes this: for some errors it will still return failed Deferreds, but other errors might be raised immediately as direct exceptions. Hence all these changes in testing code.

This actually changes current API and potentially may break some existing code. (inlineCallbacks-based user code won't be affected). 

Previous behavior can be easily saved by returning `defer.fail(...)` instead of raising, but it looks ugly.

What do you think?

psi29a:
Accepting your Chicken removal code. :)

This has been along time coming.

trenton42:
I wish I had more time to dive in and help out here. Thanks for this, @IlyaSkriblovsky 👏

trenton42:
And thanks @psi29a for reviewing 👏

PR #195, opened 2016-10-13
ConnectionPool.drop_database

IlyaSkriblovsky:
API is following PyMongo's `MongoClient.drop_database()`


psi29a:
Gladly mergable... just need changelog entry. :P

IlyaSkriblovsky:
Added one :)

Issue #55, opened 2014-04-23
Args order


claytondaley:
Any particular reason that TxMongo's args order differ from PyMongo (e.g. `find()`)?  I'm migrating some code (and have other code that I'd prefer to have working on both platforms).  Obviously, I'll update the overlapping code to use kwargs, but it was an unexpected discovery (and might catch others).

psi29a:
Since we are making changes here, I think it might be time to revisit this and get things back on track to being comparable to PyMongo.

psi29a:
Any comments on this? @fiorix

fiorix:
Well, so long as it's agreed that this change will break backwards compatibility with all existing code.

psi29a:
I would hope to document this, with BIG warnings.

The other option would be to use *args **kwargs, so that argument order wouldn't matter. It would involve a bit more code though.

fiorix:
Non-breaking changes are definitely better. Pretty sure the day you release a new version with a backwards incompatible call for such a popular function it'll break many people's env.

claytondaley:
I certainly agree with @fiorix that BC breaks are bad.  It just strikes me as really weird not to mimic PyMongo's syntax and -- if the change is ever going to be made -- the sooner the better.  

All the more so if we expect a substantial upgrade that could attract a lot of migration from PyMongo.  Leaving the current syntax both (1) creates a migration hassle for migrating PyMongo users and, once migrated, (2) increases inertia for the current syntax.

psi29a:
@claytondaley I agree with you, i rather have BC breaks that get txmongo on the right path than to let inertia hold us back.

From my point of view, we should have one big 15.0 release with all our bug fixes and new features. That way people can use that release without breakage.

The follow release, 15.1 will be the 'breakage'. This gives people the chance to pin against 15.0 until they fix their applications.

I really want to avoid having to do (if it works) *args **kargs tricks and have to carry that forward.

claytondaley:
I like the basic plan, but wouldn't sync the TxMongo versions with Twisted.  Release the updates (pre BC) as 0.7 (the only release right now is 0.6) and make the BC change 1.0.

Separately, I don't think `args kwars` will help.  The two args without keywords will be `args[0]` and `args[1]`.  `filter` and `projection` are both dicts so a quick type check won't help.  You'd be stuck doing something complex to distinguish between the two possibly orders.

psi29a:
The next release will be 15.0, we're already moved over to Twisted.

[...]

PR #196, opened 2016-10-14
New signature for find() method and compatibility with old one

IlyaSkriblovsky:
Continuing #55

I suggest to change signature of `find()`, `find_with_cursor()` and `find_one()` to more closely match PyMongo's counterparts.

Now it is `(spec=None, skip=0, limit=0, fields=None, filter=None, cursor=False, **kwargs)`
I'm proposing to change it to `(spec=None, projection=None, skip=0, limit=0, filter=None, **kwargs)`

It can be breaking if done immediately, but we can do it with deprecation interval. This PR makes `find*()` to accept both old and new signatures with issuing DeprecationWarning for the old one.

* Short form `find({'x': 1}, {'_id': 0})` becames compatible with PyMongo
* Keyworded form `find({'x': 1}, projection={'_id': 0})` too
* I suggest to deprecate `cursor` argument to `find()` in favor of `find_with_cursor()`. `cursor=True` changes behavior and semantics of `find()` completely, so it is better looking to have distinct method for this.

Further thoughts:
May be it worth to also rename `spec` and `filter` to `filter` and `sort` correspondingly?
Pros:
* `filter` becames compatible with PyMongo
* `sort` becames more like in `find_one_for_update`

Cons:
* `sort` is not technically correct term because it may be used also for hints, comments, etc.
* `find(filter={'x': 1})` will not be 100%-distinguishable between old and new variants.

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -20,6 +20,20 @@ Features
 API Changes
 ^^^^^^^^^^^
 
+- ``find()``, ``find_one()``, ``find_with_cursor()``, ``count()`` and ``distinct()`` signatures
+  changed to more closely match PyMongo's counterparts. New signatures are:
+  
+  - ``find(filter=None, projection=None, skip=0, limit=0, sort=None, **kwargs)``
+  - ``find_with_cursor(filter=None, projection=None, skip=0, limit=0, sort=None, **kwargs)``
+  - ``find_one(filter=None, projection=None, **kwargs)``
+  - ``count(filter=None, **kwargs)``
+  - ``distinct(key, filter=None, **kwargs)``
+  
+  Old signatures are now deprecated and will be supported in this and one subsequent releases. 
+  After that only new signatures will be valid.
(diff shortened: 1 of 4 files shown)


psi29a:
I'm all for this, keeping things similar with PyMongo is what we aim for anyway.

It can be confusing and potentially break things keeping the filter keyword but changing its function. I believe this is a risk worth taking to steer TxMongo back to PyMongo.

IlyaSkriblovsky:
Ok, I think this PR is good to go. I will continue to work on `Collection` methods arguments, but this particular change is completed.

I intentionally didn't changed testing code to use new signatures to ensure compatibility.

psi29a:
Hmmm, something not right with test_hint...

IlyaSkriblovsky:
Oops, seems that `count()` with invalid hint only fails loudly in 3.2. Fixed.

psi29a:
LGTM, merging!

IlyaSkriblovsky:
Cool! Plz take a look on the follow-up: #199

trenton42:
@IlyaSkriblovsky You have been on fire lately! 👏  👍

PR #199, opened 2016-10-17
kwargs removed where it is not necessary

IlyaSkriblovsky:
removed `kwargs` where it is not necessary and made `@timeout`'s `_deadline` argument explicit in most cases

### DIFF
--- a/tests/test_collection.py
+++ b/tests/test_collection.py
@@ -46,7 +46,7 @@ class TestIndexInfo(unittest.TestCase):
 
     @defer.inlineCallbacks
     def tearDown(self):
-        yield self.coll.drop()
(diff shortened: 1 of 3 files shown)


psi29a:
:+1:  Yes, less kwarg magic... thanks you

psi29a:
```
txmongo/collection.py:26: 'txmongo.protocol.INSERT_CONTINUE_ON_ERROR' imported but unused
```
hmmm

IlyaSkriblovsky:
Thanks, removed that unused import. It became unused after simplifying internal function that was before used for both `insert _one` and `insert _many` but now serves only `insert _one`.

psi29a:
merging!  :D

PR #200, opened 2016-10-21
Retrieve the number of versions of any given file

oddjobz:
This seems like missing critical functionality, once you start building versions of a given file, you really need (at some point) to be able to determine how many versions of that file you have. Cycling through previous would seems to be a horribly inefficient mechanism. This method is 'better', although maybe someone with a better grasp of txmongo's internals could improve it .. for now tho' it provides the functionality and I've added a basic test to show it works.


psi29a:
Nice...

tests/test_count.py
```diff
@@ -0,0 +1,57 @@
...
+    def test_GFSNoCount(self):
+        try:
+            count = yield self.gfs.count('worldx')
+        except NoFile:
+            self.assertEqual(1, 1)
+            return
+        except Exception:
+            raise
+
+        self.assertEqual("exception", "missing")
```
IlyaSkriblovsky: it is more idiomatic to call `self.fail("should not come here")` to intentionally fail the test

txmongo/_gridfs/__init__.py
```diff
@@ -177,6 +177,19 @@ def ok(cursor):
         return self.__files.find(query, filter=filter.sort(myorder), limit=1, skip=skip)\
             .addCallback(ok)
 
+    @defer.inlineCallbacks
+    def count(self, filename=None):
```
IlyaSkriblovsky: Why do `filename` has a default value `None`? Does `count()` return something meaningful if called without arguments?

txmongo/_gridfs/__init__.py
```diff
@@ -177,6 +177,19 @@ def ok(cursor):
...
+        """Count the number of versions of a given file.
+        Returns an integer number of versions of the file in GridFS whose filename matches
+        `filename`, or raises NoFile if the file doesn't exist.
+        :Parameters:
+          - `filename`: ``"filename"`` of the file to get version count of
+        """
+        cursor = yield self.__files.find({"filename": filename})
+        if cursor:
+            defer.returnValue(len(cursor))
+        raise NoFile("no versions for filename %r" % (filename))
```
IlyaSkriblovsky: That's strange for a function named "count" to raise an exception if there is no files. Intuition suggest that it should simply return zero. This way it will be more like same-named `Collection.count()`.

txmongo/_gridfs/__init__.py
```diff
@@ -177,6 +177,19 @@ def ok(cursor):
...
 
+    @defer.inlineCallbacks
+    def count(self, filename=None):
+        """Count the number of versions of a given file.
+        Returns an integer number of versions of the file in GridFS whose filename matches
+        `filename`, or raises NoFile if the file doesn't exist.
+        :Parameters:
+          - `filename`: ``"filename"`` of the file to get version count of
+        """
+        cursor = yield self.__files.find({"filename": filename})
```
IlyaSkriblovsky: It would be more efficient to call `self.__files.count({"filename": filename})` here. Not just because this way we won't move any file metadata across the TCP connection, but what is more important, MongoDB will be able to give us an answer only by querying index and not touching actual collection data at all.

txmongo/_gridfs/__init__.py
IlyaSkriblovsky: If you agree with returning zero, the whole function can be simplified like this:
```python
def count(self, filename):
    """ ... """
    return self.__files.count({"filename": filename})
```

oddjobz:
Ok, how's that look now?  :)

IlyaSkriblovsky:
Cool! But `test_GFSNoCount` can be simplified by removing `try..except` :)

oddjobz:
Like that?

oddjobz:
:)

IlyaSkriblovsky:
I'll merge your code when testing is finished. Thanks for contribution!

psi29a:
Thanks very much for this! :)

PR #207, opened 2017-04-26
Auth using authsource db

harlov:
Use 'authsource' as auth database, if specify in uri, and orig database if not.

https://docs.mongodb.com/manual/reference/connection-string/#urioption.authSource

### DIFF
--- a/txmongo/connection.py
+++ b/txmongo/connection.py
@@ -272,7 +272,8 @@ class ConnectionPool(object):
         ]
 
         if self.__uri['database'] and self.__uri['username'] and self.__uri['password']:
-            self.authenticate(self.__uri['database'], self.__uri['username'],
+            auth_db = self.__uri['options'].get('authsource') or self.__uri['database']
+            self.authenticate(auth_db, self.__uri['username'],
                               self.__uri['password'],
                               self.__uri['options'].get('authmechanism', 'DEFAULT'))
 
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -8,6 +8,7 @@ Bugfixes
 ^^^^^^^^
 
 - Memory leak fixed in `Collection.bulk_write()`
+- Use authSource as auth database if specify in connect uri
 
 Release 16.3.0 (2016-11-25)
 ---------------------------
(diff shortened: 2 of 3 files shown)


txmongo/connection.py
```diff
@@ -298,6 +298,10 @@ def __tcp_or_ssl_connect(self, host, port, factory, **kwargs):
             return reactor.connectTCP(host, port, factory, **kwargs)
 
     @property
+    def auth_db(self):
+        return self.__uri['options'].get('authsource') or self.__uri['database']
```
IlyaSkriblovsky: This property adds to the public API of Connection without obvious reason. Why do we need this as a public property? Or even if we need it, its name falsely hints that it will return `Database` object, not `str`. My thinking that there is no bold reason to make this public and it is better to make it local variable inside `__init__`.

txmongo/connection.py
```diff
@@ -271,8 +271,8 @@ def __init__(self, uri="mongodb://127.0.0.1:27017", pool_size=1, ssl_context_fac
             for i in range(pool_size)
         ]
 
-        if self.__uri['database'] and self.__uri['username'] and self.__uri['password']:
-            self.authenticate(self.__uri['database'], self.__uri['username'],
+        if self.__uri['database'] and self.__uri['username'] and self.__uri['password']:
```
IlyaSkriblovsky: Sorry for tediousness, but could you please remove the trailing whitespace?

IlyaSkriblovsky:
Thanks for your contribution! Could you please address a couple of comments above?

txmongo/connection.py
harlov: Sorry) i will fix.

txmongo/connection.py
harlov: Yes, property reduntant. I will fix it.

txmongo/connection.py
harlov: done.

txmongo/connection.py
harlov: done

IlyaSkriblovsky:
Thank you!

PR #211, opened 2017-08-10
Fixed compatibility with pymongo 3.5.0 (#210)

IlyaSkriblovsky:

Issue #210, opened 2017-08-10
Pymongo 3.5.0 doesn't provide _WriteOp in pymongo.operations


bhyde:
This line:

https://github.com/twisted/txmongo/blob/c960defdb8190d2e66aaea88c13e163498b1784c/txmongo/collection.py#L17

 is failing since recent release of pymongo 3.5.0.

Pinning to 3.4.0 is a work around.

bhyde:
See also: https://github.com/mongodb/mongo-python-driver/commit/9468c119a31b5809dc5883605b5be91eb4bc57ac#diff-e03bf8345fe8fe509a1d13a5e75edb26

psi29a:
@bhyde Thank you for reporting this!

Looks like we only import _WriteOp to validate that the bulk_insert request is of instance _WriteOp. If I understand correctly, pymongo instead just does a try/except around request._add_to_bulk(blk) in which case we can do the same.

@IlyaSkriblovsky Do you agree?

IlyaSkriblovsky:
Sure, fixed in #211
Thanks, Ben!

psi29a:
@IlyaSkriblovsky I was looking for consensus that my assumption was correct, not that you do the work! ;)

IlyaSkriblovsky:
Sorry, it's sometimes hard to resist when fingers reach the keyboard :)

psi29a:
Hah! No worries!

bhyde:
Oh, nice turn around.  Awesome.  I'll keep an eye out for a release.   Thanks!

psi29a:
Merged... going to close this. If there is a still an issue, please re-open.

Cheers!

PR #212, opened 2017-08-10
Tests fixed for MongoDB 3.4

IlyaSkriblovsky:
I've noticed that several tests were failing against modern MongoDB 3.4

1. https://github.com/twisted/txmongo/compare/master...IlyaSkriblovsky:mongodb-3.4-tests-fixed?expand=1#diff-cecc483f704c4be15eaed66c3bb3678bL302
MongoDB 3.4 [changed behavior of the first batch of `find()` command](https://docs.mongodb.com/v3.4/tutorial/iterate-a-cursor/#cursor-batches). Before it was returning maximum of 1mb in the first batch, now this restriction is loosened, so this test fails. Actually I think this test is redundant, it was checking that negative limit forces only the single batch in result. It is not our concern anyways.

2. https://github.com/twisted/txmongo/compare/master...IlyaSkriblovsky:mongodb-3.4-tests-fixed?expand=1#diff-cecc483f704c4be15eaed66c3bb3678bL384
MongoDB 3.4 changed error message in that case

3. https://github.com/twisted/txmongo/compare/master...IlyaSkriblovsky:mongodb-3.4-tests-fixed?expand=1#diff-1f3700c74d32647bfd888de73a25b21dL1025
This one is actually a bug. Ghost `_deadline` kwarg were added in index specification by `create_index`. MongoDB ≤3.2 just ignores it, so it wasn't noticed. 3.4 throws an error here.

### DIFF
--- a/tests/test_queries.py
+++ b/tests/test_queries.py
@@ -298,12 +298,6 @@ class TestLimit(SingleCollectionTest):
         res = yield self.coll.find(limit=-150)
         self.assertEqual(len(res), 150)
 
-    @defer.inlineCallbacks
-    def test_HardLimitAboveMessageSizeThreshold(self):
-        yield self.coll.insert([{'v': ' '*(2**20)} for _ in range(8)], safe=True)
-        res = yield self.coll.find(limit=-6)
-        self.assertEqual(len(res), 4)
-
 
 class TestSkip(SingleCollectionTest):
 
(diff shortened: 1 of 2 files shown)


IlyaSkriblovsky:
Test will probably fail due to #210. I will rebase this PR onto #211 once it merged.

psi29a:
Good catch on # 3, 3.4 does validation checking on indexes now and we caught this in another area of our code at work where MongoEngine was using safe=True and type=False. So it make sense that it would choke on _deadline.

psi29a:
Hmmm:

>  This job ran on our Precise environment, which is in the process of being decommissioned. Please read about the rollout of Trusty as default on the blog, and take note that you can explicitly stay on Precise by specifying dist: precise in your .tr

Guess I'll look into this as well.

IlyaSkriblovsky:
Seems like Travis is still testing against MongoDB 3.0. Can we fix this too as well? :)

psi29a:
I'll get on that this weekend.

PR #213, opened 2017-08-11
17.1.0

IlyaSkriblovsky:
I believe that quick fix from #211 deserves a release because simple `pip install txmongo` results in broken installation right now (since it installs txmongo 16.3.0 and pymongo 3.5.0 which are not compatible).

Should we name the new release 17.1.0? Or may be 16.3.1 because it consists only of bugfixes?

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -1,14 +1,15 @@
 Changelog
 =========
 
-Release 17.1.0 (UNRELEASED)
+Release 17.1.0 (2017-08-11)
 ---------------------------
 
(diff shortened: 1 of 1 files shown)


psi29a:
Go ahead for 17.1.0... we have enough material.

IlyaSkriblovsky:
So, could you please merge this if you don't have objections and publish the new release to PyPI?

psi29a:
pushed... 

https://pypi.python.org/pypi/txmongo

IlyaSkriblovsky:
Thanks!

Issue #218, opened 2017-12-07
Incompotable witn pymongo 3.6.0


tonal:
insert_one error:
```
2017-12-07 12:28:50 [scrapy.utils.signal] ERROR: Error caught on signal handler: <bound method MongodbSaver.spider_opened of <remains.mongo_ext.MongodbSaver object at 0x7f6a10d28208>>
Traceback (most recent call last):
  File "/home/tonal/lang/projects/remains/venv3/lib/python3.6/site-packages/twisted/internet/defer.py", line 1384, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "/home/tonal/lang/projects/remains/venv3/lib/python3.6/site-packages/twisted/python/failure.py", line 408, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "/home/tonal/lang/projects/remains/remains/mongo_ext.py", line 56, in spider_opened
    ins_ret = yield tasks.insert_one(doc)
  File "/home/tonal/lang/projects/remains/venv3/lib/python3.6/site-packages/twisted/internet/defer.py", line 653, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/home/tonal/lang/projects/remains/venv3/lib/python3.6/site-packages/txmongo/collection.py", line 635, in on_ok
    _check_write_command_response([[0, response]])
  File "/home/tonal/lang/projects/remains/venv3/lib/python3.6/site-packages/pymongo/helpers.py", line 206, in _check_write_command_response
    write_errors = result.get("writeErrors")
AttributeError: 'list' object has no attribute 'get'
```
replace_one error:
```
2017-12-07 12:29:25 [scrapy.utils.signal] ERROR: Error caught on signal handler: <bound method MongodbSaver.spider_closed of <remains.mongo_ext.MongodbSaver object at 0x7f6a10d28208>>
Traceback (most recent call last):
  File "/home/tonal/lang/projects/remains/venv3/lib/python3.6/site-packages/twisted/internet/defer.py", line 1384, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "/home/tonal/lang/projects/remains/venv3/lib/python3.6/site-packages/twisted/python/failure.py", line 408, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "/home/tonal/lang/projects/remains/remains/mongo_ext.py", line 84, in spider_closed
[...]

PR #219, opened 2017-12-07
Fixing compatibility with pymongo 3.6

IlyaSkriblovsky:
PyMongo 3.6 changed behavior of private `pymongo.helpers._check_write_command_response()` function. I've copied it into `txmongo.collection` to avoid future compatibility issues.

See #218

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+Release 17.2.0 (UNRELEASED)
+---------------------------
+
+Bugfixes
+^^^^^^^^
+
+- Fixed compatibility with PyMongo 3.6
+
 Release 17.1.0 (2017-08-11)
 ---------------------------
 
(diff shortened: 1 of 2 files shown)


psi29a:
Looks good, waiting for tests to run. 

Should we also add MongoDB 3.6 as default, maybe not in this PR but in another?

psi29a:
Seems to be a problem with MongoDB too...

IlyaSkriblovsky:
Looks like it's something on Travis' side because our testing code not even started, these sleeps are part of preparation code

IlyaSkriblovsky:
It seems like Travis jobs is now run on trusty instead of precise and now testing hosts have MongoDB 3.2 preinstalled. Previously our testing jobs ran against MongoDB 3.0.
And somehow it causes MongoDB to be inaccessible while testing, don't know why.

Bret, is it possible to have some interactive terminal in the Travis' testing environment? Or may be you have any ideas what's going on?

psi29a:
Should we default to 3.6?

Sadly there is no interactive terminal, it is a bit trial and error-ish.

IlyaSkriblovsky:
@psi29a I haven't check txmongo on 3.6 yet. Hope it will work :)

PR #221, opened 2017-12-31
Fixing and cleaning Travis CI configuration

IlyaSkriblovsky:
I tried to fix Travis CI issue noticed in #219, but went a bit too far :)

(this PR includes changes from #219 in order to pass tests with PyMongo 3.6)

Seems like updated Travis' `trusty` environment doesn't starts newly-installed mongodb-org-server anymore. So I've added manual `mongod` run in `before_install`.

Then build started to fail with tox wasn't able to find Python 3.3... I have no idea of what have changed on the Travis' side, but anyways our previous Travis config seems a bit strange because it ran all py2.7–3.5 tests with `python: 2.7` in .travis.yml... So I decided to make it a bit more clear.

• Manually starting mongod since updated Travis' `trusty` image doesn't starts it after installation anymore
• More clean testing matrix configuration
• Removed testing on Python 3.3 since PyCrypto dropped support for 3.3 and doesn't compiles on it anymore
• Added testing on pypy3
• PyCrypto doesn't compile on PyPy because it tries to build it's fastmath native extension which is incompatible with PyPy. PyCrypto builds this extension only if `libgmp-dev` is installed, so I added the line that uninstalls it on PyPy builds.
• Added newer Twisted versions for testing against

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,78 +1,73 @@
 language: python
-python: 2.7
+python:
+  - "2.7"
+  - "3.4"
+  - "3.5"
+  - "pypy"
+  - "pypy3"
 
-# container infra, when fixed
-#sudo: false
-#
-# addons:
-#  apt:
-#    sources:
-#    - mongodb-3.0-precise
-#    - pypy
-#addons:
-#  apt:
-#    packages:
-#    - mongodb-org-server
-#    - pypy
 
 env:
-    - TOX_ENV=py27-tw140
-    - TOX_ENV=py27-tw150
-    - TOX_ENV=py27-tw155
-    - TOX_ENV=py27-tw160
-    - TOX_ENV=py27-tw163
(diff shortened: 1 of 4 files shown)


psi29a:
LGTM... shouldn't you be enjoying the new year? ;)

psi29a:
Thank you for putting the effort into this. I like that we are a bit more explicit here.

IlyaSkriblovsky:
> LGTM... shouldn't you be enjoying the new year? ;)
What could be better than finally resolving the issue and finally deleting reminder that was snoozed a dozen of times during a weeks? :)

IlyaSkriblovsky:
I think fixed PyMongo 3.6 compatibility deserves a release because right now txmongo installed by `pip install txmongo` is broken. I will prepare the version bump PR.

PR #225, opened 2018-02-23
PEP8 - WIP

trenton42:
This is a work in progress, but I wanted to make sure nothing in here was going to be a problem before progressing through other things. Let me know if you have any feedback!

### DIFF
--- a/txmongo/collection.py
+++ b/txmongo/collection.py
@@ -1,6 +1,7 @@
-# Copyright 2009-2015 The TxMongo Developers.  All rights reserved.
-# Use of this source code is governed by the Apache License that can be
-# found in the LICENSE file.
+"""Copyright 2009-2018 The TxMongo Developers.  All rights reserved.
+Use of this source code is governed by the Apache License that can be
+found in the LICENSE file."""
+# pylint: disable=W0622
 
 from __future__ import absolute_import, division
 import io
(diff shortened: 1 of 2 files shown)


IlyaSkriblovsky:
Why did you added `else` in line 230, but deleted it in line 292? :) I don't mind, just curious what variant is considered to be a Good Practice™? I personally like `else` for its explicitness, but don't like for its redundancy, so I'm okay with both.

psi29a:
Honestly, I don't care for the 80 column line limit. Can we at least set this 120? 

:)

IlyaSkriblovsky:
> Honestly, I don't care for the 80 column line limit. Can we at least set this 120?

yes-yes-yes

trenton42:
> Honestly, I don't care for the 80 column line limit. Can we at least set this 120?

I 100% agree as well. If I did trim to 80, it was on accident. I have the 80 column line limit turned off in my PEP8 extension, and instead use the 120 limit from pylint.

trenton42:
@IlyaSkriblovsky Hmm, it looks like I removed it from both 🤔

This is from pylint's R1705:

> Unnecessary “else” after “return” Used in order to highlight an unnecessary block of code following an if containing a return statement. As such, it will warn when it encounters an else following a chain of ifs, all of them containing a return statement.

I tend to prefer it, but if you think it goes too far, I can revert.

IlyaSkriblovsky:
@trenton42, ouch, sorry, you're right

txmongo/collection.py
```diff
@@ -91,7 +94,9 @@ def __init__(self, database, name, write_concern=None, codec_options=None):
             msg = "TxMongo: collection names must not contain '$', '{0}'".format(repr(name))
             raise InvalidName(msg)
         if name[0] == "." or name[-1] == ".":
-            msg = "TxMongo: collection names must not start or end with '.', '{0}'".format(repr(name))
+            msg = "TxMongo: collection names must not start or end with '.', '{0}'".format(
+                repr(name)
+            )
```
psi29a: Shouldn't this fit on one line with 120?

trenton42: Yes, I think I might have gotten over aggressive on those. I will fix them.

PR #231, opened 2018-05-07
Fixing KeyError when bulk operation is interrupted (for example by failover)

IlyaSkriblovsky:
If Bulk operation is interrupted (by failover, by network failure, ...) current version of txmongo with combination of PyMongo 3.6.0 raises KeyError from `coll.bulk_write`. This is because we are calling `pymongo.bulk._merge_command` without checking it for being proper bulk response.

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -2,6 +2,14 @@ Changelog
 =========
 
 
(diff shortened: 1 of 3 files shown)


psi29a:
Great, just an update in the change log and we're good to go.

Issue #232, opened 2018-07-05
TxMongo: command SON(...) failed with 'not authorized' on remains to execute command


tonal:
After upgrade pymongo to 3.7.0 break TxMongo
```
2018-07-05 03:30:03,714 [scrapy.utils.signal] ERROR: Error caught on signal handler: <bound method MongodbSaver.spider_opened of <remains.mongo_ext.MongodbSaver object at 0x7fad95940dd8>>
Traceback (most recent call last):
  File "/opt/pyenv/versions/3.6.6/envs/remains/lib/python3.6/site-packages/twisted/internet/defer.py", line 1384, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "/opt/pyenv/versions/3.6.6/envs/remains/lib/python3.6/site-packages/twisted/python/failure.py", line 422, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "/opt/remains_grab/remains/remains/mongo_ext.py", line 56, in spider_opened
    item_cnt=self.items_scraped, page_count=self.response_count))
  File "/opt/pyenv/versions/3.6.6/envs/remains/lib/python3.6/site-packages/twisted/internet/defer.py", line 653, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/opt/pyenv/versions/3.6.6/envs/remains/lib/python3.6/site-packages/txmongo/database.py", line 67, in on_ok
    _check_command_response(response, msg, allowable_errors)
  File "/opt/pyenv/versions/3.6.6/envs/remains/lib/python3.6/site-packages/pymongo/helpers.py", line 155, in _check_command_response
    raise OperationFailure(msg % errmsg, code, response)
pymongo.errors.OperationFailure: TxMongo: command SON([('insert', 'tasks'), ('documents', [{'_id': ObjectId('5b3d668b7f7d193791c668c4'), 'type': 'site', 'name': 'Акваполис', 'load_start': datetime.datetime(2018, 7, 5, 3, 30, 3, 658310), 'load_end': datetime.datetime(2018, 7, 5, 3, 30, 3, 684830), 'status': 'start', 'msg': None, 'item_cnt': 0, 'page_count': 0}]), ('ordered', True), ('writeConcern', {})]) on namespace remains.$cmd failed with 'not authorized on remains to execute command { insert: "tasks", ordered: true, writeConcern: {}, $db: "remains" }'
```
[...]

PR #233, opened 2018-07-05
Compatibility with PyMongo 3.7

IlyaSkriblovsky:
Fixing #232 

Copy-pasted some internal pymongo utility methods into txmongo so future pymongo updates won't break txmongo.

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -10,6 +10,7 @@ Bugfixes
 
 - In combination with PyMongo 3.6.0 `bulk_write` might sometimes raise
   KeyError when bulk operation was interrupted (by failover, for example)
+- Compatibility with PyMongo 3.7
 
 
 Release 18.1.0 (2018-03-21)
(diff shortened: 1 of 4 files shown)


IlyaSkriblovsky:
I've also noticed that some of our tests are failing with MongoDB 3.6. Will look into it too.

psi29a:
Should we start keeping documentation stating where we lifted code from PyMongo? I can imagine that in the future that something breaks that was 'fixed' in a later version of PyMongo. 

In comment:
- reason why we copied with link to issue
- which pymongo version, maybe even link to the tagged released file/line

What do you think?

IlyaSkriblovsky:
I think this is great idea. Going to add this, thanks.

psi29a:
That's good enough, that we know it's copied can give us a clue.

IlyaSkriblovsky:
If this looks good I think we should make a release because currently `pip install txmongo` is installing non-working environment.

PR #234, opened 2018-07-06
Tests fixed with MongoDB 3.6

IlyaSkriblovsky:
Fixing some test failures when running against MongoDB 3.6. TxMongo works fine with 3.6, this issue only affects testing code that hardcodes some error message texts and other internal MongoDB stuff from 3.4. No user-visible changes here.

This PR is based on #233 and includes its commits, so it should be merged only after that.

### DIFF
--- a/tests/mongod.py
+++ b/tests/mongod.py
@@ -68,7 +68,6 @@ class Mongod(object):
                 b"--noprealloc", b"--nojournal",
                 b"--smallfiles", b"--nssize", b"1",
                 b"--oplogSize", b"1",
-                b"--nohttpinterface",
         ]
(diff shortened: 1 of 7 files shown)


tests/mongod.py
```diff
@@ -68,7 +68,6 @@ def start(self):
                 b"--noprealloc", b"--nojournal",
                 b"--smallfiles", b"--nssize", b"1",
                 b"--oplogSize", b"1",
-                b"--nohttpinterface",
```
psi29a: Thank you for this! I had to do this for an internal project too. :)

psi29a:
Done and done.

PR #240, opened 2018-10-31
Allow collection options to be passed only via kwargs

stevenkaras:
This is in line with the pymongo [behavior][0], and will make it easier to migrate code from pymongo to txmongo.

[0]: https://github.com/mongodb/mongo-python-driver/blob/3.7.2/pymongo/database.py#L304

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+Release 18.3.0 (UNRELEASED)
+---------------------------
+
+Features
+^^^^^^^^
+
+- Allow passing only kwargs to `Database.create_collection()`
+
 
 Release 18.2.0 (2018-07-19)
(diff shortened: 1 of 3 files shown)


IlyaSkriblovsky:
I think this is the very good idea to be more in line with pymongo. Thanks!
/cc @psi29a

psi29a:
Agreed!

psi29a:
@stevenkaras can you add an entry here:
https://github.com/twisted/txmongo/blob/master/docs/source/NEWS.rst
Something like:

> Release 18.3.0 (UNRELEASED)
> * Feature: <your entry here>

Once we have that, we'll merge! :)

Thank you very much1

stevenkaras:
@psi29a is there anything else?

psi29a:
Good to go! Thanks. :)

Issue #247, opened 2019-03-24
do not connect to mongodb+srv:// scheme


tonal:
From https://github.com/twisted/txmongo/blob/master/txmongo/connection.py#L254
```python
        if not uri.startswith("mongodb://"):
            uri = "mongodb://" + uri

        self.__uri = parse_uri(uri)
```
Bat  pymongo.uri_parser.parse_uri support it.

IlyaSkriblovsky:
Thanks for reporting!
I've created a quick PR in #248.

PR #248, opened 2019-03-25
support for mongodb+srv:// uri scheme

IlyaSkriblovsky:
As reported in #247 we didn't support monogdb+srv:// although PyMongo supports it since 3.6.

This PR adds mongodb+srv scheme with "extra" `[srv]` dependency because corresponding feature in PyMongo requires same extra dependency too.

So to have mongodb+srv working txmongo should be installed like this: `pip install txmongo[srv]`.


psi29a:
lgtm!  Thanks

IlyaSkriblovsky:
Sorry, I've forgot to update changelog as usual :)  See #249

IlyaSkriblovsky:
By the way, we have some unreleased features, I think it worth to make a release now

PR #256, opened 2019-08-21
Pymongo 3.9 and mongodb 4.0

IlyaSkriblovsky:
Seems like Travis updated their workers to MongoDB 4.0, so we won't be able to run tests until both PyMongo 3.9 and MongoDB 4.0 compatibility issues are fixed. So combining #254 and #255 into this single PR.


psi29a:
This looks reasonable. Thanks! :)

PR #257, opened 2019-08-21
Mongodb 4.2 compatibility

IlyaSkriblovsky:
• Fixes `create_index()` on MongoDB 4.2 by using `createIndexes` command instead of inserting into `system.indexes` collection. This method is compatible with at least MongoDB 2.6+, so this change should be safe.
• Fixes tests for MongoDB 4.2

(This PR incorporates #256 and should be probably merged after that)


psi29a:
I do not see anything obviously wrong, thanks!

Issue #252, opened 2019-08-15
builtins.AttributeError: '_CaseInsensitiveDictionary' object has no attribute 'copy'


tonal:
versions:
txmongo            19.1.0
pymongo            3.9.0

Error in txmongo/connection.py line: 259 ConnectionPool.__init__
```
wc_options = self.__uri['options'].copy()
```
Forget `copy()` in `pymongo._CaseInsensitiveDictionary`

IlyaSkriblovsky:
Thanks for reporting. PyMongo <3.9 used plain dict there...

userlond:
Temporary fix is to downgrade to pymongo 3.8.0

tonal:
@userlond I use monkeypath:
```python
import pymongo
if pymongo.version_tuple == (3, 9, 0):
  def _cid_copy(self):
    return self.__class__(**self)
  pymongo.common._CaseInsensitiveDictionary.copy = _cid_copy
```

PR #258, opened 2019-09-16
19.2.0 release

IlyaSkriblovsky:
I think we need to release because currently `pip install txmongo` results in broken installation due to #252

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -1,7 +1,7 @@
 Changelog
 =========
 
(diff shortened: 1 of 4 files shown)


psi29a:
On this today.

psi29a:
Uploaded

IlyaSkriblovsky:
Thank you!

I'm preparing important update to support sessions and transactions. So I hope it won't be long to wait for 19.3.0 :)

PR #262, opened 2020-04-02
new-style wire protocol implementation for find*() methods

IlyaSkriblovsky:
Starting from maxWireVersion=4 (v3.2) MongoDB migrated to new implementation of CRUD methods based on [commands](https://docs.mongodb.com/manual/reference/command/nav-crud/) commands instead of `OP_*` low-level message codes.

We already support new-style commands for insert, update, replace and delete (when called as `*_one`/`*_many`). This patch adds support for find*() methods too and uses it by default if maxWireVersion ≥ 4.

This also (hopefully) makes txmongo compatible with Azure's Cosmos DB that only supports new-style API.

In future this will allow us to support sessions and transactions because it only supported for new-style queries.

### DIFF
--- a/tests/test_filters.py
+++ b/tests/test_filters.py
@@ -17,6 +17,7 @@ from __future__ import absolute_import, division
 
 from twisted.trial import unittest
 from twisted.internet import defer
+
 import txmongo
 import txmongo.filter as qf
 from pymongo.errors import OperationFailure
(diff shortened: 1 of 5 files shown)


psi29a:
Still want to move this forward?

IlyaSkriblovsky:
No, this will be superseded by OP_MSG implementation on which we are working here: https://github.com/IlyaSkriblovsky/txmongo/tree/op-msg-support

PR #274, opened 2021-06-03
Fix retrieving error messages

### LINKED ISSUE #236: "no SNI name sent" when connecting to a Mongo Atlas M0 Cluster
I am trying to connect to an Atlas M0 (Free Tier) Cluster with txmongo.

Following the instructions here: https://github.com/twisted/txmongo/issues/203 I created the connection pool with:
`db = yield ConnectionPool(url, ssl_context_factory=ssl.ClientContextFactory())`

But I get the following error in the logs when trying to start my app:
```
2018-08-10 16:14:43-0400 [MongoProtocol (TLSMemoryBIOProtocol),client] 'TxMongo: Unknown error'
2018-08-10 16:14:44-0400 [MongoProtocol (TLSMemoryBIOProtocol),client] 'TxMongo: Unknown error'
2018-08-10 16:14:44-0400 [MongoProtocol (TLSMemoryBIOProtocol),client] 'TxMongo: Unknown error'
2018-08-10 16:14:44-0400 [MongoProtocol (TLSMemoryBIOProtocol),client] <twisted.internet.tcp.Connector object at 0x114b6ec18> will retry in 2 seconds
```

Here is more info about actual error (as seen in `txmongo/connection.py l105`):

`config = {'ok': 0, 'errmsg': 'no SNI name sent, make sure using a MongoDB 3.4+ driver/shell.', 'code': 8000, 'codeName': 'AtlasError'}`

Do you know how I can solve this ?

Thank you for your time.

```
Config:

MacOS 10.13.6
MongoDB shell version v3.6.6
Python 3.6.4
pymongo==3.7.1
txmongo==18.2.0
Twisted==18.7.0
```

Gowee:
Hi.

As is reported in #236, error messages in exceptions raised by txmongo are sometimes ambiguous.

The reason is that txmongo only picks the field `err`, which is not present all the time even if there are actual errors. I cannot find a detailed and exhaustive documentation for all possible fields (there are some related documentations, but for example, none of them formally documents the existence of the field `codeName`). But at least, there are `codeName`, and `errmsg` in recent MongoDB versions.

This PR is going to fix the problem by fall backing to `codeName` and `errmsg` when `err` is not present. It picks any of the three fields whichever is present. 

It should partially fix #236, giving a detailed error message:
`TxMongo: AtlasError: no SNI name sent, make sure using a MongoDB 3.4+ driver/shell.`

### DIFF
--- a/txmongo/connection.py
+++ b/txmongo/connection.py
@@ -15,7 +15,7 @@ from twisted.python import log
 from twisted.python.compat import StringType
 from txmongo.database import Database
 from txmongo.protocol import MongoProtocol, Query
-from txmongo.utils import timeout
+from txmongo.utils import timeout, get_err
 
 DEFAULT_MAX_BSON_SIZE = 16777216
 DEFAULT_MAX_WRITE_BATCH_SIZE = 1000
@@ -103,7 +103,7 @@ class _Connection(ReconnectingClientFactory):
         # Make sure the command was successful.
         if not config.get("ok"):
             code = config.get("code")
-            msg = "TxMongo: " + config.get("err", "Unknown error")
+            msg = "TxMongo: " + get_err(config, "Unknown error")
             raise OperationFailure(msg, code)
 
         # Check that the replicaSet matches.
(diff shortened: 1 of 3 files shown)


IlyaSkriblovsky:
Thanks, this looks reasonable

PR #288, opened 2024-09-16
Deprecation warnings

IlyaSkriblovsky:
I want to make TxMongo compatible with newer wire protocol of MongoDB 5.0+ (OP_MSG).

This would be much easier to do without supporting older protocols at the same time.

That's why I suggest to drop support of outdated and no longer supported version of Python and MongoDB. This will allow us to cleanup and simplify the codebase.

To do it nicely I suggest to release 23.0.0 with the only change of deprecation warnings when old Python or MongoDB are detected.

@psi29a what do you think about it?

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -1,6 +1,16 @@
 Changelog
 =========
 
+Release UPCOMING (yyyy-mm-dd)
+-----------------------------
+
+API Changes
+^^^^^^^^^^^
+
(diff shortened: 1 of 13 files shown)


IlyaSkriblovsky:
By the way, seems like we broke py2.7 compatibility already almost couple years ago by this line:
https://github.com/twisted/txmongo/blob/master/txmongo/collection.py#L8

IlyaSkriblovsky:
I've rebased this PR onto #290

psi29a (approved):

PR #294, opened 2024-10-14
OP_MSG support

IlyaSkriblovsky:
* New [MongoDB wire protocol](https://github.com/mongodb/specifications/blob/master/source/message/OP_MSG.md) implemented. It is mandatory to support MongoDB 5.0 and higher.
* Tons of legacy code removed since MongoDB compatibility bumped to 4.0+
* Old-style find* methods signature removed


txmongo/utils/__init__.py
```diff
@@ -57,7 +58,7 @@ def on_fail(failure):
     return _timeout
 
 
-def check_deadline(_deadline):
+def check_deadline(_deadline: Optional[float]):
```
psi29a, line 61: Should we switch away from `Optional[float]` and just use `float | None`; and typing in general? Or keep that change for another PR?

psi29a (approved):

txmongo/utils/__init__.py
IlyaSkriblovsky: Would love to, but `T | S` syntax is only for 3.10+ and `list[T]` instead of `List[T]` is 3.9+

PR #300, opened 2024-11-04
Builder pattern methods on Cursor

IlyaSkriblovsky:
Add builder-pattern methods for Cursor allowing doing queries like this:

```python
async for doc in collection.find({"size": "L"}).sort({"price": 1}).skip(10).limit(50):
    print(doc)
```


NikolayBaranovv (approved):

psi29a (approved):
Awesome!

PR #308, opened 2025-03-25
Fix compatibility with Twisted 24.10+

IlyaSkriblovsky:
Current version of TxMongo is broken when used with Twisted ≥24.10.0 because we forgot to add `"result"` to `_DEFERRED_METHODS` :(

Code to reproduce:

```python
@react
async def main(reactor):
    conn = ConnectionPool()
    db = conn.db
    coll = db.coll

    await coll.drop()
    await coll.insert_one({"x": 1})

    print(await coll.find())

    await conn.disconnect()
```

This example hangs hardly on `await coll.find()`

### DIFF
--- a/docs/source/NEWS.rst
+++ b/docs/source/NEWS.rst
@@ -25,6 +25,11 @@ API Changes
 
 - find_one_and_* methods now returns None if called with unacknowledged write_concern.
 
+Bugfixes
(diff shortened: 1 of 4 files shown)


psi29a (approved):
Glad it was found

PR #309, opened 2025-03-30
25.0.0

IlyaSkriblovsky:
Let's release 25.0.0!


psi29a (approved):
Let's go! So much good stuff here.

